PR #1, opened 2026-04-07
feat: integrate QMD semantic search into wiki-ingest and wiki-query skills

hongphuc5497:
## Summary

  - **wiki-ingest**: Added QMD semantic search integration to check existing wiki pages before creating new ones, preventing duplicate content and enabling smarter
  incremental ingestion
  - **wiki-query**: Integrated `mcp__qmd__query` as the primary search mechanism, with multi-strategy search (lex + vec) and minScore filtering for high-confidence results

  ## Motivation

  The QMD MCP server provides fast local semantic search over compiled wiki pages. Integrating it into the two most-used skills enables:
  - **Deduplication during ingest** — agents check for existing pages on a topic before creating new ones, then update instead of duplicating
  - **Better query answers** — wiki-query uses vector + keyword search to find the most relevant pages rather than relying on file traversal

  ## Changes

  | File | Change |
  |---|---|
  | `.skills/wiki-ingest/SKILL.md` | Added QMD pre-check step before page creation |
  | `.skills/wiki-query/SKILL.md` | Replaced manual file search with `mcp__qmd__query` multi-strategy retrieval |

  ## Test plan

  - [ ] Run wiki-ingest on a document already partially covered — verify it updates rather than duplicates
  - [ ] Run wiki-query on a topic in the wiki — verify cited, relevant results via QMD
  - [ ] Run wiki-query on a topic not in the wiki — verify graceful fallback

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,7 +29,7 @@ LINT_SCHEDULE=weekly
 # --- QMD Semantic Search (optional) ---
 #
 # QMD is a local MCP server that indexes your wiki and sources for fast semantic search.
-# Install: https://github.com/Ar9av/qmd
+# Install: https://github.com/tobi/qmd
 # After installing, point these at your QMD collection names (set when you ran `qmd index`).
 #
 # Without QMD: wiki-ingest and wiki-query fall back to Grep/Glob — fully functional, just slower.
--- a/.env.example
+++ b/.env.example
@@ -25,3 +25,26 @@ CLAUDE_HISTORY_PATH=
 
 # Lint schedule: daily | weekly | manual
 LINT_SCHEDULE=weekly
+
+# --- QMD Semantic Search (optional) ---
+#
+# QMD is a local MCP server that indexes your wiki and sources for fast semantic search.
+# Install: https://github.com/Ar9av/qmd
+# After installing, point these at your QMD collection names (set when you ran `qmd index`).
+#
+# Without QMD: wiki-ingest and wiki-query fall back to Grep/Glob — fully functional, just slower.
+# With QMD: semantic search replaces grep passes, enabling concept-level matches and cross-collection queries.
+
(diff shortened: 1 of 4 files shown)


Ar9av (changes requested):
hi, i tested this. I think at times the "skip" branch harder to misread. The fallback prose is good but a defensive agent will still sometimes try MCP u have mcp__qmd__query with an empty  collection name. 

You can consider a one-line guard at the top of each QMD step: "If $QMD_WIKI_COLLECTION is empty or unset, do not call mcp__qmd__* at all — jump 
  to Step N." smth like this

hongphuc5497:
> hi, i tested this. I think at times the "skip" branch harder to misread. The fallback prose is good but a defensive agent will still sometimes try MCP u have mcp__qmd__query with an empty collection name.
> 
> You can consider a one-line guard at the top of each QMD step: "If $QMD_WIKI_COLLECTION is empty or unset, do not call mcp__qmd__* at all — jump to Step N." smth like this

@Ar9av Thanks for the feedback! You're spot on — I see what you mean. An agent could read through the fallback notes and still try to hit the MCP endpoint even with an empty collection name. I've added a guard check right at the top of each step:

```md
  wiki-ingest, Step 1b:
  **GUARD: If `$QMD_PAPERS_COLLECTION` is empty or unset, skip this entire step and proceed to Step 2.**

  wiki-query, Step 2b:
  **GUARD: If `$QMD_WIKI_COLLECTION` is empty or unset, skip this entire step and proceed to Step 3.**
```

This way, before an agent even reads the rest of the step or tries anything, it checks: is the collection set? If not, just skip to the next step. Everything else stays the same — all the fallback prose and logic is still there for when QMD is configured. Just added this explicit gate at the front so there's no ambiguity. Pushed in `abe43e1`. Let me know if that works better!

Ar9av (approved):
lgtm, thanks

PR #3, opened 2026-04-07
Summary parser safe punctuation

Medenor:
## Problem

  The `wiki-update` skill’s **Page format** template showed `summary:` as an unquoted YAML scalar.
  When generated summaries include `:` (common in technical writing), frontmatter can parse incorrectly in Obsidian/YAML tooling.

  ## What Changed

  - Updated the **Page format** template to use folded scalar style for summary:
    - `summary: >-`
    - indented summary content on the next line
  - Added explicit guidance below the template:
    - Prefer folded scalar (`>-`) for `summary` to avoid YAML parsing issues with punctuation.
    - Keep summary content indented by two spaces under `summary: >-`.
    - This avoids quote-escaping rules while remaining parser-safe.
  - Updated the instruction sentence to require folded scalar summary format for all new/updated pages.

  ## Why

  This removes an avoidable YAML footgun and makes generated frontmatter more robust across punctuation-heavy summaries without needing escaping
  logic.

  ## Scope

  Documentation/template-only change in `wiki-update/SKILL.md` (no runtime/tooling code changes).

  ## Validation

  - Confirm the template now renders `summary` using folded style (`>-`).
  - Confirm guidance is present and unambiguous.
  - Manual check: summaries containing `:`, `#`, and quotes parse correctly as frontmatter with folded style.

### DIFF
--- a/.skills/wiki-update/SKILL.md
+++ b/.skills/wiki-update/SKILL.md
@@ -104,7 +104,7 @@ title: Page Title
 category: concepts
 tags: [tag1, tag2]
 sources: [projects/<project-name>]
-summary: One or two sentences (≤200 chars) describing what this page covers.
+summary: "One or two sentences (≤200 chars) describing what this page covers."
 provenance:
   extracted: 0.6
   inferred: 0.35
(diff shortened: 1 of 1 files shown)


Ar9av (approved):
lgtm , thanks a lot

PR #6, opened 2026-04-11
refactor(setup): extract install_skills() to eliminate duplicated symlink logic

Medenor:
The symlink installation loop (guard check → remove stale link → skip real
directories → ln -s) was copy-pasted four times across steps 2, 3b, 3c, and
3d with only the target directory varying. This makes bugs in the logic require
four separate fixes and obscures what each section is actually doing.

Extract the loop into a reusable `install_skills <target_dir> <label>` function
defined at the top of the script. Each call site is now a single line, and the
guard logic lives in exactly one place.

Changes:
- Add install_skills() function with local variables (target_dir, label,
skill_name, link_path) to avoid polluting the global shell scope
- Replace step 2's 18-line inner for-loop with a single install_skills call
- Replace steps 3b, 3c, and 3d (three identical ~15-line blocks) with three
install_skills calls
- Inline the GLOBAL_SKILLS array literal since it was only used once
- Replace `ls -d … | wc -l | tr -d ' '` skill count with a glob + grep -c
pipeline that avoids parsing ls output

Net: -47 lines (-78/+31), that is, 24% reduction.

Behavior is unchanged.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -20,6 +20,27 @@ set -e
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 SKILLS_DIR="$SCRIPT_DIR/.skills"
 
+# Symlink every skill in SKILLS_DIR into TARGET_DIR.
+# Skips real directories to avoid data loss; updates stale symlinks.
+install_skills() {
+  local target_dir="$1"
+  local label="$2"
+  mkdir -p "$target_dir"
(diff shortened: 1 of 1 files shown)


Ar9av (approved):

Ar9av (approved):
Yeah man, makes sense 
Thanks for the improvement

PR #9, opened 2026-04-12
fix(setup): handle vault paths with spaces correctly

Medenor:
## Summary
Paths with spaces (e.g. `/LLMs Brain`) were not properly escaped or quoted during setup, causing some LLMs to create parallel directories instead of finding the correct vault location.

## Changes
- Quote vault paths in `.env` to preserve spaces during parsing
- Properly escape special characters (/, &, ") when updating `.env` with sed
- Quote paths in the generated `~/.obsidian-wiki/config` file

## Test Plan
- [x] Setup script correctly handles vault paths with spaces
- [x] Existing paths without spaces continue to work
- [x] Config file is readable by skills

## Why this matters
Skills that read the vault path from `~/.obsidian-wiki/config` now work reliably regardless of spaces in the directory name.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -64,7 +64,8 @@ mkdir -p "$GLOBAL_CONFIG_DIR"
 # Read vault path from .env if it's already set
 VAULT_PATH=""
 if [ -f "$SCRIPT_DIR/.env" ]; then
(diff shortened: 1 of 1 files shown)


Ar9av (approved):
lgtm

Issue #30, opened 2026-05-06
Make config resolution consistent across skills


madebymlai:
Several skills resolve wiki configuration differently, which makes non-global or multi-vault setups fragile.

Current examples:

- `wiki-ingest`, `data-ingest`, `cross-linker`, etc. use:
  ```text
  ~/.obsidian-wiki/config (preferred) or .env (fallback)
  ```

- `wiki-rebuild`, `wiki-status`, `wiki-lint`, history ingest skills use:
  ```text
  .env
  ```

- `wiki-query` says:
  ```text
  ~/.obsidian-wiki/config ... Fall back to .env if you're inside the obsidian-wiki repo
  ```

- `wiki-update`, `daily-update`, and `memory-bridge` require or strongly assume:
  ```text
  ~/.obsidian-wiki/config
  ```

- `daily-update` also writes global state:
  ```text
  ~/.obsidian-wiki/.last_update
  ~/.obsidian-wiki/.pending_delta
  ```

This creates drift between skills and makes setups with subject-local/project-local `.env` files harder to use. For example, a university notes workspace may have one wiki per subject:

```text
vault/ADE/.env
vault/ADE/wiki/
vault/MDL/.env
vault/MDL/wiki/
```

In that setup, a global `~/.obsidian-wiki/config` is either wrong or ambiguous.

Could the project standardize config loading across skills?

Suggested behavior:

1. Prefer the nearest `.env` found by walking up from the current working directory.
2. If no local `.env` is found, fall back to `~/.obsidian-wiki/config`.
3. If neither exists, ask the user or run setup.
4. Skills that write state, such as `daily-update`, should write state per vault or per resolved config, not only to global `~/.obsidian-wiki/*`.
5. Ideally this should be implemented as one shared config resolver/helper, not copied instructions in every skill. New skills should not read `~/.obsidian-wiki/config` or `.env` directly; they should use the same resolution contract.
6. Document this as the shared config-resolution contract so new skills do not drift.

This would preserve the current single-global-wiki workflow while also supporting multi-vault and subject/project-local workflows.

madebymlai:
One related drift point: `daily-update` setup mode also appends to `~/.zshrc`:

```sh
echo "" >> ~/.zshrc
echo "# obsidian-wiki terminal notification" >> ~/.zshrc
echo "source $OBSIDIAN_WIKI_REPO/scripts/wiki-notify.sh" >> ~/.zshrc
```

[...]

PR #31, opened 2026-05-06
docs: sweep remaining config protocol references

madebymlai:
## Summary

Follow-up sweep for #30 after the shared Config Resolution Protocol landed in `llm-wiki`.

This updates the remaining direct skill entry points and repo-level agent docs that still described the old config order (`~/.obsidian-wiki/config` preferred, `.env` fallback, or `.env` only). They now point agents at the shared protocol instead:

```text
walk up CWD for .env -> ~/.obsidian-wiki/config -> prompt setup
```

Also updates the `llm-wiki` link-format note to say `OBSIDIAN_LINK_FORMAT` comes from the resolved config, rather than naming config files directly.

## Validation

- `git diff --check`
- `rg` sweep for stale config-resolution wording across `.skills`, `AGENTS.md`, README, setup docs, and scripts returns no matches

### DIFF
--- a/.skills/cross-linker/SKILL.md
+++ b/.skills/cross-linker/SKILL.md
@@ -18,7 +18,7 @@ You are weaving the wiki's knowledge graph tighter by finding and inserting miss
 
 ## Before You Start
 
-1. Read `~/.obsidian-wiki/config` (preferred) or `.env` (fallback) to get `OBSIDIAN_VAULT_PATH` and `OBSIDIAN_LINK_FORMAT` (default: `wikilink`)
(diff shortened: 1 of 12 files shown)


Ar9av:
Reviewed — clean follow-up to #30. Verified `Config Resolution Protocol` exists in `.skills/llm-wiki/SKILL.md` and ripgrep confirms no stale `~/.obsidian-wiki/config (preferred) or .env (fallback)` wording remains across `.skills/`, `AGENTS.md`, and `README.md`. Doc-only, all skill entry points now point at the shared protocol consistently. LGTM.

PR #33, opened 2026-05-08
feat: add wiki-switch skill for multi-vault profile management

nova28:
## Summary

- Adds a new `wiki-switch` skill that lets users manage multiple named Obsidian vault profiles
- Active vault is controlled via a symlink at `~/.obsidian-wiki/config` pointing to `~/.obsidian-wiki/config.NAME`
- Supports four operations: **switch** (activate a vault), **list** (show all vaults + active), **show** (print config, redacting secrets), and **new** (scaffold a new vault config from the active one)
- Routing row added to `AGENTS.md`

## Usage

| Command | Description |
|---|---|
| `/wiki-switch NAME` | Switch to a named vault |
| `/wiki-switch list` | List all vaults and show which is active |
| `/wiki-switch show [NAME]` | Print config for a vault (secrets redacted) |
| `/wiki-switch new NAME` | Scaffold a new vault config from the active one |

### DIFF
--- a/.skills/wiki-switch/SKILL.md
+++ b/.skills/wiki-switch/SKILL.md
@@ -1,0 +1,98 @@
+---
+name: wiki-switch
+description: >
+  Switch between multiple Obsidian wiki vault profiles. Use this skill when the user says
+  "/wiki-switch NAME", "switch to my work wiki", "switch vault", "change wiki", "which wiki am I on",
+  "list my wikis", "show my vaults", "create a new vault config", or "add a new wiki profile".
+  The skill manages named config files at ~/.obsidian-wiki/config.NAME and activates one by
+  symlinking it to ~/.obsidian-wiki/config.
+---
+
+# Wiki Switch — Manage Multiple Vault Profiles
+
+Each vault is a complete config file at `~/.obsidian-wiki/config.<name>`. The active vault is
+whichever file `~/.obsidian-wiki/config` symlinks to. Switching vaults means re-pointing that symlink.
+
+## Dispatch
+
+Parse the invocation and route to the right section:
+
+| Invocation | Action |
+|---|---|
+| `/wiki-switch <name>` | → **Switch** |
+| `/wiki-switch list` | → **List** |
+| `/wiki-switch show [name]` | → **Show** |
+| `/wiki-switch new <name>` | → **New** |
+| `/wiki-switch` (no args) | → **List** (treat as list) |
+
+---
+
+## Switch (default action)
+
+Activate a named vault profile.
+
+1. Verify `~/.obsidian-wiki/config.<name>` exists. If not, tell the user the vault doesn't exist and list what's available (run **List**).
+2. Run:
+   ```bash
+   ln -sf ~/.obsidian-wiki/config.<name> ~/.obsidian-wiki/config
+   ```
+3. Read `OBSIDIAN_VAULT_PATH` from the newly active config.
+4. Confirm to the user:
+   ```
+   Switched to vault: <name>
+   Vault path: <value of OBSIDIAN_VAULT_PATH from the config>
+   ```
+
+---
+
+## List
+
+Show all registered vault profiles and which is active.
+
+1. Find all files matching `~/.obsidian-wiki/config.*` (exclude `config` itself — that's the symlink).
+2. Resolve the current symlink target: `readlink ~/.obsidian-wiki/config`
+3. For each config file, read the first non-empty comment line (lines starting with `#`) as a human description of the vault. Fall back to the file's suffix as the label if no comment exists.
+4. Display:
+   ```
+   Vaults:
+     personal   My personal research wiki    ← active
+     work       Work projects wiki
+   ```
+   Mark the active one with `← active`. If the symlink is broken or `config` doesn't exist, show `(none active)`.
+
+---
+
+## Show
+
+Print the full config for a vault.
+
+- If a name is given, read `~/.obsidian-wiki/config.<name>`.
+- If no name given, read `~/.obsidian-wiki/config` (the active vault).
+- If the file doesn't exist, tell the user and list what's available.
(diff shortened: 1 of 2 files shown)


Ar9av (commented):
## Review: wiki-switch skill

Good addition overall — multi-vault switching fills a real gap for users with work/personal wikis. The symlink-based approach is clean and non-destructive. A few things to address in a follow-up:

### Blocker: Missing IDE mirrors
Every skill in this repo is mirrored to four locations:
- `.claude/skills/<name>`
- `.cursor/skills/<name>`
- `.windsurf/skills/<name>`
- `.agents/skills/<name>`

This PR only adds `.skills/wiki-switch/SKILL.md`. The four IDE mirrors are missing — commit `23c4a92` was literally a fix for the same gap. Please sync the skill to all four mirrors (can copy the file verbatim).

### Minor: `CLAUDE.md` routing table not updated
`CLAUDE.md` has its own skill routing table that mirrors the one in `AGENTS.md`. The `wiki-switch` row needs to be added there too, otherwise the Claude Code agent won't route to it.

### Suggestion: Cross-reference `wiki-setup`
The `new` operation creates a second vault config, but a user setting up a brand-new vault from scratch still needs `wiki-setup`. Worth adding a one-liner cross-reference so they know to run `wiki-setup` after `wiki-switch new`.

Merging now — the mirrors + CLAUDE.md update can land as a follow-up.

Ar9av:
Follow-up items from this review have been tracked in #34.

myroslav:
@Ar9av symlink-based solution is "global". It does not allow simultaneous use of one wiki in one project and other wiki in the other project. 

IMHO, per-project config (`$PROJECT_ROOT/.obsidian-wiki/config`) would work better.

nova28:
Great point @myroslav — you're right that the symlink approach forces a single global active vault, which breaks down if you're context-switching between projects with different wikis.

One direction I've been thinking about: instead of just switching *to* a vault, what if the system supported **multi-vault querying** — where you can compose vaults together for a session?

For example:
- `personal + work` — query both, consolidate the answer
- `personal + investment` — pull from personal context and investment notes together

The active vault would still handle writes, but reads could fan out across a declared set. Something like a `OBSIDIAN_VAULT_QUERY_PATH` list in config, or a session-level flag like `/wiki-query --vaults personal,investment`.

Not sure this is the right solution — it adds complexity and the consolidation logic needs thought. But it feels closer to how people actually use multiple knowledge bases (they overlap, not just alternate).

Would love to hear your thoughts @myroslav @Ar9av — is per-project config the simpler win here, or is multi-vault querying worth exploring?

nova28:
The `wiki-switch` skill we just merged is a pragmatic first step — but myroslav's point made me think about what the real end goal looks like.

What I actually want is **cross-vault querying**: compose two or more vaults together for a query (`personal + investment`, `personal + work`) and get a consolidated answer with citations per vault. That's the feature that would truly solve the problem. But that's non-trivial — fan-out reads, dedup, consolidation logic across vaults. So I stepped back to the simpler solution first.

---

## What already works

The `projects/` directory already handles most separation needs — `$VAULT/projects/<name>.md` scopes knowledge within a single vault without needing a second vault at all. For most "work vs personal" use cases, this is enough.

## When do you actually need a separate vault?

Not for project separation — `projects/` handles that. A separate vault makes sense when you need:
- **Different audience** — a kids' wiki should never bleed into a personal search
- **Hard privacy boundary** — investment notes you don't want appearing in unrelated queries
- **Different taxonomy/conventions** — each domain has its own tags, writing style, structure

## What's genuinely missing

**1. Per-project config** — myroslav's suggestion of `$PROJECT_ROOT/.obsidian-wiki/config` is the right pattern. The current config resolution supports a global `~/.obsidian-wiki/config` or walking up for `.env` — but `.env` files carry project secrets and mixing wiki config there is wrong. A dedicated `.obsidian-wiki/config` at the project root, mirroring the global config format, would let the active vault switch automatically based on where you're working. This is new code, not something that exists today.

**2. Cross-vault querying** — once you have multiple vaults, you can't query across them. Proposal — named vault groups in `~/.obsidian-wiki/groups`:
```
research   = personal, investment
everything = personal, work, investment
```
`/wiki-query --group research "what do I know about SaaS metrics"` fans out across all vaults in the group, consolidates the answer, and cites the source vault per result.

**Writes always go to the single active vault** — groups are read-only compositions. No ambiguity about where new knowledge lands.

**3. Ingest routing (nice-to-have)** — default to the active vault at ingest time. Power users can override via config rules:
```
ingest_routes:
  investment/* → investment
  health/*     → personal
  *            → active vault
```

## User journeys

**First-time user:** `wiki-setup` → one vault, done. Never sees any of this.

**Growing user:** Use `projects/` first — only create a second vault when you hit a real audience or privacy boundary. `wiki-switch` handles the manual switching in the meantime.

**Multi-vault power user:** `$PROJECT_ROOT/.obsidian-wiki/config` for auto-switching per project, vault groups for cross-vault queries.

---

So the natural progression:
1. **Now** — `wiki-switch` (manual switching, merged in #33)
2. **Next** — per-project config (`$PROJECT_ROOT/.obsidian-wiki/config`) for auto-switching
3. **Later** — cross-vault querying with vault groups

Does that roadmap make sense @myroslav @Ar9av?

myroslav:
At the moment Obsidian Vault at `OBSIDIAN_VAULT_PATH` provides 2-level structure: global (level 1) and per-project (level 2). This is suitable for a shallow knowledge organization model (for instance for a single multirepo software project).

For more complex setups, when there is deeper herarchy, or multiple multirepo software projects (multiple obsidian-wiki vaults) only "primary" vault (`.env`: `OBSIDIAN_VAULT_PATH`) will probably work for 80% of the cases, but ability to utilize knowledge from the other multirepo software projects, or from higher-up in the knowledge hierarchy would be great.

PR #42, opened 2026-05-12
Add QMD CLI transport support

BenRoe:
## Summary
- add `QMD_TRANSPORT`, `QMD_CLI_SEARCH_MODE`, and `QMD_CLI` config options
- document MCP vs CLI QMD usage in `.env.example`, README, and setup guidance
- teach `wiki-ingest` and `wiki-query` how to use the local `qmd` CLI when MCP is unavailable
- make `wiki-query` load QMD settings from the resolved config and prefer QMD before grep when configured
- document keeping operator-like terms in `lex:` and using natural language for `vec:` queries

## Why CLI mode
- lower agent token usage: the agent can call `qmd` directly and consume compact ranked results instead of loading broader wiki content first
- simpler setup for agents without MCP support: no MCP server wiring is required when `qmd` is already installed locally
- fewer runtime dependencies: CLI mode relies on the existing `qmd` binary and configured collections rather than an agent-specific MCP integration

## Test
- ran a local `qmd query` smoke test from outside the vault using `QMD_TRANSPORT=cli`-style settings
- verified the lex+vec query returned ranked wiki files before any page-body grep was needed
- confirmed a hyphenated `vec:` term can be parsed as negation and that rewriting the vector query in plain language succeeds

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,7 +37,7 @@ OBSIDIAN_LINK_FORMAT=wikilink
 
 # --- QMD Semantic Search (optional) ---
 #
-# QMD is a local MCP server that indexes your wiki and sources for fast semantic search.
+# QMD indexes your wiki and sources for fast semantic search.
 # Install: https://github.com/tobi/qmd
 # After installing, point these at your QMD collection names (set when you ran `qmd index`).
 #
@@ -49,6 +49,20 @@ QMD_WIKI_COLLECTION=
 
 # Name of the QMD collection indexing your raw source documents (OBSIDIAN_SOURCES_DIR or _raw/)
 QMD_PAPERS_COLLECTION=
+
+# QMD integration transport: mcp | cli
+# mcp preserves the original behavior and uses an agent-configured QMD MCP server.
+# cli runs the local qmd command directly; useful for agents without QMD MCP support.
+QMD_TRANSPORT=mcp
+
+# CLI search quality mode: quality | balanced | fast
+# quality: qmd query with reranking (best relevance, slowest on CPU)
+# balanced: qmd query --no-rerank (good hybrid results, faster)
+# fast: qmd vsearch (semantic only) or qmd search when exact terms matter
+QMD_CLI_SEARCH_MODE=quality
+
+# Optional qmd binary override if qmd is not on PATH.
+QMD_CLI=qmd
 
 # --- Raw Staging Directory (optional) ---
 #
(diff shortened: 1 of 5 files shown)


BenRoe:
Follow-up note on the `vec:` guidance:

QMD structured queries treat `-term` as an exclusion operator. That is fine for lexical/operator-style search, but `vec`/`hyde` queries do not support negation. This means a natural-looking hyphenated phrase can fail if it contains a `-term` segment.

Example that can fail:

```text
vec: dotfiles no-sudo user-local CLI role patterns
```

QMD may parse the `-sudo` part as negation and return an error like:

```text
Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.
```

Safer pattern:

```text
lex: dotfiles no-sudo user-local CLI ~/.local/bin ansible_become=false
vec: dotfiles user local CLI role patterns and installs without sudo
```

So the rule of thumb is: keep exact punctuation-heavy tokens in `lex:`, and make `vec:` plain natural language.

Ar9av (commented):
Reviewed the full diff. Here are some inline observations:

**`.env.example`**
Defaulting `QMD_TRANSPORT=mcp` is the right backwards-compat call — existing setups with QMD MCP configured just keep working with zero changes. The `QMD_CLI=qmd` default also cleanly handles the common case where `qmd` is already on PATH.

**`wiki-query/SKILL.md` — vec: negation warning**
This is a genuinely useful piece of defensive guidance that's easy to miss:
> *QMD treats `-term` as negation, and negation is not supported in `vec`/`hyde` queries.*

The concrete examples (`no-sudo`, `ansible_become=false`, `~/.local/bin`) make it very actionable. Operators with hyphens or special chars → `lex:`, plain prose → `vec:`. Clear.

**`wiki-query/SKILL.md` — QMD eagerness**
The shift from "run QMD when the index pass didn't produce clear candidates" to "run QMD before Grep unless already answered by hot.md/index.md" is a meaningful improvement. For semantic or cross-project questions, the old heuristic would skip QMD too often. The new phrasing treats QMD as the first-class retrieval path when configured, which matches the intent of having it installed.

**Fallback chain**
The consistent "if transport unavailable, skip QMD and proceed to Step 3" pattern across both `wiki-ingest` and `wiki-query` is solid. It means misconfigured transport (qmd not on PATH, MCP not wired) degrades gracefully to grep rather than erroring mid-retrieval.

**Step renumbering in wiki-query**
Steps 1–3 → 1–4 is handled cleanly. No dangling references to old step numbers spotted.

Overall the changes are well-scoped, backwards-compatible, and the documentation quality is high.

Ar9av (approved):
Good job — clean, well-documented, and backwards-compatible. The CLI transport is a real quality-of-life win for setups without MCP, and the vec: negation caveat is exactly the kind of gotcha that belongs in the skill doc. Merging.

PR #52, opened 2026-05-15
fix(setup): use relative symlinks for in-repo skill mirrors

tarekrached:
## Summary

`setup.sh` writes **absolute** symlinks (e.g. `/Users/<name>/obsidian-wiki/.skills/<skill>`) into the in-repo skill mirror directories (`.agents/`, `.claude/`, `.cursor/`, `.kiro/`, `.windsurf/`), but those mirrors are tracked in git as **relative** symlinks (`../../.skills/<skill>`). Every fresh clone is left with ~160 modified files immediately after running `setup.sh`.

This is the same class of bug that 476ac88 (*"complete relative symlink migration across all mirrors and skills"*) fixed in the committed tree — that migration updated the symlinks themselves but missed the installer that recreates them. A contributor running `git add -A` after setup would silently re-commit their home directory into the public repo, undoing 476ac88.

### Repro on `main`

```
$ git clone https://github.com/Ar9av/obsidian-wiki && cd obsidian-wiki
$ bash setup.sh
$ git status --short | wc -l
     160
$ git diff .claude/skills/claude-history-ingest
-../../.skills/claude-history-ingest
+/Users/<your-name>/obsidian-wiki/.skills/claude-history-ingest
```

### Fix

Add an optional third `mode` argument to `install_skills()`. When `"relative"` is passed, the link target is `../../.skills/<name>` — correct for every in-repo mirror, all of which live at `$REPO/<agent-dir>/skills`. Global target dirs under `$HOME` keep the absolute target, which is what they need.

Also fixes a misleading log line — it previously said *"Installed global skills"* even for in-repo dirs.

## Test plan

- [x] Fresh clone + `bash setup.sh` → `git status` is clean
- [x] `.claude/skills/<name>` → `../../.skills/<name>` (relative, matches committed)
- [x] `~/.codex/skills/<name>` → `/abs/path/to/.skills/<name>` (absolute, unchanged)
- [x] Re-running `setup.sh` on an already-installed clone is idempotent (still clean)

### DIFF
--- a/.github/workflows/setup.yml
+++ b/.github/workflows/setup.yml
@@ -1,0 +1,29 @@
+name: setup
+
+# Guards setup.sh against the regression where it rewrites committed
+# relative symlinks with absolute paths (or any other modification to
+# tracked files). See #52 for the original bug.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  install-leaves-tree-clean:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run setup.sh
+        # stdin closed so the vault-path prompt returns empty rather than hang
+        run: bash setup.sh < /dev/null
+      - name: Working tree must be unchanged
+        run: |
+          if ! git diff --exit-code; then
+            echo "::error::setup.sh modified tracked files. Likely cause: an" \
+                 "in-repo symlink mirror is being written with an absolute" \
+                 "target instead of a relative one. See install_skills() in setup.sh."
+            exit 1
+          fi
+      - name: Re-run is idempotent
+        run: bash setup.sh < /dev/null && git diff --exit-code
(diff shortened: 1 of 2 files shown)


Ar9av:
## Test Results ✅

Tested on this machine (Ubuntu, not a fresh clone — pre-existing `.gitignore` modification unrelated to this PR).

### What was verified

| Check | Result |
|---|---|
| `bash setup.sh < /dev/null` completes without error | ✅ |
| In-repo mirror symlinks are relative (`../../.skills/<name>`) | ✅ |
| Global dir symlinks are absolute (`/abs/path/.skills/<name>`) | ✅ |
| Re-running `setup.sh` is idempotent | ✅ |
| Only pre-existing `.gitignore` change in working tree (not caused by setup.sh) | ✅ |

### Symlink spot-check

```
.claude/skills/wiki-query  → ../../.skills/wiki-query   ✅ relative
.cursor/skills/wiki-query  → ../../.skills/wiki-query   ✅ relative
.agents/skills/wiki-query  → ../../.skills/wiki-query   ✅ relative
~/.claude/skills/wiki-query → /abs/path/.skills/wiki-query  ✅ absolute
~/.codex/skills/wiki-query  → /abs/path/.skills/wiki-query  ✅ absolute
```

The fix correctly discriminates between in-repo mirrors (relative mode) and global `$HOME`-based dirs (absolute mode). The new CI workflow (`.github/workflows/setup.yml`) will guard against regression on every future push.

LGTM — ready to merge.

PR #75, opened 2026-05-29
fix(wiki-ingest): derive promoted page sources from _raw frontmatter, not _raw path

BenRoe:
## Problem

When promoting `_raw/` drafts in Raw Mode, the skill instructed agents
to "include the source in the `sources` frontmatter field" with no
guidance on which source to use. Agents defaulted to the `_raw/<filename>`
path — a staging artifact — instead of the real origin.

`wiki-quick-chat-capture` already writes two relevant frontmatter fields
into every `_raw/` file:
- `capture_source: claude-session` — who wrote the draft
- `sources: "<project> session (<YYYY-MM-DD>)"` — what session it came from

The ingest step was silently discarding both and recording the internal
staging path instead.

## Fix

Two edits to `.skills/wiki-ingest/SKILL.md`:

1. **Raw Mode section** — added a `Source inheritance:` rule with a
   priority ladder:
   - Both `capture_source` + `sources` present → synthesize combined entry:
     `"agent:<capture_source> source:<capture_source> project:<project> date:<lifecycle_changed>"`
   - Only `sources` present → copy verbatim
   - Neither present → fall back to `_raw/` filename

2. **Step 5 "Include the source" bullet** — inline reminder pointing to
   the Raw Mode rule so the instruction is visible at the point of action.

## Why this matters

`wiki-quick-chat-capture` (and any future capture tool) writes structured
provenance into `_raw/` files precisely so promoted pages carry accurate
origin data. The ingest step must not discard it.

### DIFF
--- a/.skills/wiki-ingest/SKILL.md
+++ b/.skills/wiki-ingest/SKILL.md
@@ -64,6 +64,13 @@ Process draft pages from the `_raw/` staging directory inside the vault. Use whe
 
 In raw mode, each file in `OBSIDIAN_VAULT_PATH/_raw/` (or `OBSIDIAN_RAW_DIR`) is treated as a source. After promoting a file to a proper wiki page, **delete the original from `_raw/`**. Never leave promoted files in `_raw/` — they'll be double-processed on the next run.
 
+**Source inheritance:** The `_raw/` path is a staging artifact — never use it as the `sources:` value on the promoted page. Derive the source entry from the `_raw/` file's own frontmatter instead:
+
+- If the file has both `capture_source` and `sources:` fields, synthesize a combined entry:
+  `"agent:<capture_source> source:<capture_source> project:<project> date:<lifecycle_changed>"`
+- If the file has only `sources:`, copy those entries verbatim.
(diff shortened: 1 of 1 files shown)


Ar9av (commented):
**Review finding — bug in synthesized source format**

In the new *Source inheritance* rule, the combined-entry template has an error:

```
"agent:<capture_source> source:<capture_source> project:<project> date:<lifecycle_changed>"
```

**Problems:**
1. `source:<capture_source>` copies the same value as `agent:` — it should reference the `sources:` field value from the `_raw/` frontmatter, not repeat `capture_source`.
2. `<project>` and `<lifecycle_changed>` are undefined tokens. The `sources:` field already encodes the project and date as `"<project> session (<YYYY-MM-DD>)"`, so there's no need to decompose and re-compose them.

**Suggested fix** — keep the synthesized format simpler and correct:

```
"agent:<capture_source> <sources-value>"
```

e.g. if `capture_source: claude-session` and `sources: "obsidian-wiki session (2026-05-29)"`, the result would be:
```
"agent:claude-session obsidian-wiki session (2026-05-29)"
```

Fixing this before merge.

PR #92, opened 2026-06-12
feat(skills): add opt-in Paper Deep-Dive template for academic papers

wwwadx:
## Problem

Ingested academic papers produce thin, text-only wiki pages. For ML/AI/LLM/VLM papers the substance lives in the **architecture figure, the core equations, and the results table**, but the current pipeline drops all of it:

- the generic page template (`llm-wiki/SKILL.md`) only has `## Key Ideas` bullets / `## Open Questions` / `## Sources` — no slot for method, results, equations, or figures;
- `wiki-ingest` reads PDFs as **text by page range**, and the one vision path fires only for *image-only* PDFs, so a normal arXiv PDF's figures are never looked at;
- "distill, don't summarize" + "aim for 10–15 small pages" scatters a paper into stubs.

Net result: equations flattened into backtick code spans, results as comma-separated number blobs, zero diagrams.

## Approach — opt-in, papers only

A richer path **scoped to academic papers landing in `references/`**, leaving the terse default untouched for every other source (so it respects the project's "distill, don't summarize" philosophy):

- **`llm-wiki/SKILL.md`** — a new **Paper Deep-Dive Template** alongside the generic one: `> [!tldr]` callout, `## Problem & Motivation`, `## Method / Architecture` (real figure), `## Key Equations` as `$$…$$` LaTeX, a `## Results` table **+ a result/motivating figure**, `## Limitations`. Frontmatter/provenance/confidence/relationships unchanged.
- **`wiki-ingest/SKILL.md`** — an **Academic papers** sub-step that re-reads figure/equation-dense pages with vision and embeds the paper's **real figures**, with a concrete PyMuPDF recipe:
  - raster figures → `get_image_info(xrefs=True)` + `extract_image(xref)` (use the native `img["ext"]`, downscale big ones);
  - vector figures → `get_pixmap(clip=rect, …)` where `rect` is a drawings-only, column/caption-bounded box;
  - **Mermaid is kept as the zero-dependency fallback** when no figure can be extracted, so the contribution adds no *required* dependency. Step 5 branches to the new template for papers.
- **`ingest-prompts.md`** — a **Paper Extraction Frame** reading checklist (problem / method / equations / results / ablations).

## Design notes

- **Backward compatible.** No behavior change for non-paper sources; the generic template and existing pages stay valid.
- **No new required dependency.** Mermaid (Obsidian-native) is the baseline; PyMuPDF is only needed for the optional real-figure path.
- Reconstructed/redrawn content is flagged `^[inferred]`, consistent with the provenance convention.

## Test plan

Docs/instructions only (no code paths). Dogfooded end-to-end on three papers:
- **Coconut** (arXiv:2412.06769) — raster figure via `extract_image`, LaTeX equations, results table.
- **Qwen2-VL** (arXiv:2409.12191) — fresh ingest; raster architecture figure + scaling/capability **result figures**.
- **PIPO** (arXiv:2605.27255) — **vector** figures rendered via `get_pixmap` clip; equation upgraded from code span to LaTeX.

Verified nested code fences render, no markdown-lint CI affected, README has no stale template reference.

### DIFF
--- a/.skills/llm-wiki/SKILL.md
+++ b/.skills/llm-wiki/SKILL.md
@@ -48,7 +48,7 @@ Organize pages into these default categories (customizable in `.env`):
 | `concepts/` | Ideas, theories, mental models | `concepts/transformer-architecture.md` |
 | `entities/` | People, orgs, tools, projects | `entities/andrej-karpathy.md` |
 | `skills/` | How-to knowledge, procedures | `skills/fine-tuning-llms.md` |
-| `references/` | Summaries of specific sources | `references/attention-is-all-you-need.md` |
+| `references/` | Summaries of specific sources; academic papers use the Paper Deep-Dive Template (below) | `references/attention-is-all-you-need.md` |
 | `synthesis/` | Cross-cutting analysis across sources | `synthesis/scaling-laws-debate.md` |
 | `journal/` | Timestamped observations, session logs | `journal/2024-03-15.md` |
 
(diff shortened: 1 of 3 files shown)


Ar9av:
One inconsistency worth fixing before merge:

`ingest-prompts.md` Paper Extraction Frame Q2 says "Sketch it as a Mermaid flowchart", but `wiki-ingest/SKILL.md` positions Mermaid as the fallback, not the default. 

An LLM following the extraction checklist will draw Mermaid instead of attempting figure extraction. Q2 should say something like: "Which figure shows the architecture/pipeline? Embed it from the PDF, fall back to Mermaid if extraction fails."

The opt-in scoping, PyMuPDF recipe, and nested fence usage all check out.

Ar9av:
Tested template routing with a non-interactive claude run against arXiv:2412.06769 (Coconut). Claude correctly picked the Paper Deep-Dive Template and cited the right SKILL.md sections — the routing from Step 1 "Academic papers" through to the Step 5 template override works.

The inconsistency in the previous comment still stands: `ingest-prompts.md` Q2 says "Sketch it as a Mermaid flowchart" while `wiki-ingest/SKILL.md` labels Mermaid the fallback. The routing test hit the right path, but a model reading the extraction checklist in isolation would default to Mermaid instead of attempting figure extraction first.

wwwadx:
Good catch — fixed in 9696692.

Q2 was a leftover from before the real-figures-first switch. It now reads: embed the architecture figure from the PDF (per the *Academic papers* recipe in `wiki-ingest/SKILL.md` — extract the raster, or render the vector region), and fall back to a Mermaid flowchart only if extraction fails. So a model reading the extraction checklist in isolation now defaults to extraction, consistent with the SKILL.md fallback framing.

I also re-audited the other Mermaid mentions across the three files — the rest already say "fallback," so Q2 was the only stale one.

Thanks for the thorough review and the routing test 🙏

Ar9av:
Changes already applied to main via cherry-pick in 306c0c9 (released as v2026.06.3). Closing.

Issue #82, opened 2026-06-10
fix(stop-hook): re-fires with cumulative stats after wiki-quick-chat-capture runs


Ar9av:
## Problem

The Stop hook counts tool uses across the **entire session**, not since the last capture. After running `/wiki-quick-chat-capture`, any further activity re-triggers the hook with inflated numbers that include already-captured work.

Example: session has 8 edits and 48 shell calls. User runs `/wiki-quick-chat-capture`. Two more edits happen. Hook fires again showing 10 edits — but 8 of those were already captured.

## Root Cause

`wiki-stop-capture.sh` iterates all JSONL entries in the transcript without any cutoff point. There's no concept of "what was already captured".

## Fix

1. After `/wiki-quick-chat-capture` stages its files, `touch ~/.obsidian-wiki/capture-checkpoint`.
2. In the hook, if this file exists, use its `mtime` as a cutoff — only count transcript entries with `timestamp > mtime`.
3. If no checkpoint exists (first capture), count all entries as today.

This is session-safe: a checkpoint from a previous session will be older than all current-session entries, so the first trigger in a new session still counts everything correctly.

## Files to change

- `.claude/hooks/wiki-stop-capture.sh` — add mtime-based cutoff to the Python counter
- `.skills/wiki-quick-chat-capture/SKILL.md` — add `touch ~/.obsidian-wiki/capture-checkpoint` to Step 5

PR #93, opened 2026-06-12
fix(stop-hook): fire at most once per session via session_id sentinel

### LINKED ISSUE #91: Stop hook still counts cumulative session stats — #82 closed as "completed" but the fix never landed
## Summary

`wiki-stop-capture.sh` counts Write/Edit/Bash tool uses across the **entire
session transcript** with no cutoff. Once a session crosses the threshold
(≥1 edit or ≥4 shell calls), the Stop hook fires on **every** subsequent
turn — repeatedly nudging for `/wiki-capture --quick` even when nothing new
has happened since the last capture.

This is the exact bug described in #82. #82 is closed as **completed**, but
the fix it specifies was never committed.

## Evidence the fix never landed

- Pickaxe finds no trace of the proposed mechanism in any branch/commit:
  ```
  $ git log --oneline -S 'capture-checkpoint' --all -- .
  (no output)
  ```
- The hook has only two commits, neither adding checkpoint/mtime logic:
  ```
  $ git log --oneline --follow -- .claude/hooks/wiki-stop-capture.sh
  f057168 refactor(skills): fold 3 redundant skills into survivors (38→35) (#85)
  de3a090 feat(capture): auto-capture findings at session end via Stop hook (#81)
  ```
- #85's only change to the hook is a command-name rename
  (`/wiki-quick-chat-capture` → `/wiki-capture --quick`) — no logic change.
- `wiki-quick-chat-capture/SKILL.md` (where #82 said the `touch` should live)
  never contained any `touch`/checkpoint/mtime line in any version before it
  was deleted in #85.

Current counter (no cutoff):
```python
for line in f:
    ...
    if name in ("Write", "Edit", "NotebookEdit"): write_edit += 1
    elif name == "Bash": bash_count += 1
```

## Repro

1. New session; make ≥1 edit (crosses threshold).
2. Run `/wiki-capture --quick`; findings staged.
3. Continue the session.
4. Every subsequent turn re-fires the nudge with cumulative counts. Observed
   ~10 identical nudges in one session.

## Proposed fix

**Option A — implement #82 as written**, but keep the checkpoint write *inside
the hook* (and/or `wiki-capture` Step 7) so it isn't coupled to a skill that
can be renamed/removed:
- on fire (or on capture) `touch ~/.obsidian-wiki/capture-checkpoint`
- count only transcript entries with `timestamp > checkpoint mtime`
- no checkpoint ⇒ count all (first capture / new session)

**Option B — fire at most once per session** via a `session_id` sentinel in
the hook. Self-contained, no cross-file dependency:
```bash
# after the stop_hook_active guard
SESSION_ID=$(printf '%s' "$INPUT" | python3 -c "
import json,sys; print(json.load(sys.stdin).get('session_id',''))" 2>/dev/null || echo "")
SENTINEL=""
if [[ -n "$SESSION_ID" ]]; then
  SENTINEL="${TMPDIR:-/tmp}/wiki-stop-capture-${SESSION_ID}.done"
  [[ -f "$SENTINEL" ]] && exit 0
fi
# at the trigger, before emitting the nudge:
[[ -n "$SENTINEL" ]] && : > "$SENTINEL" 2>/dev/null || true
```

Trade-off: A re-nudges if genuinely new work appears after a capture; B
guarantees one nudge per session. B is simpler and can't be broken by skill
refactors (which is arguably how #82 ended up closed-but-unfixed).

Refs: #82 (closed completed, fix absent), #81 (hook introduced), #85 (renamed
the command, no logic change).

Ar9av:
## Summary

- Extracts `session_id` from the Stop hook payload
- Derives a sentinel path `$TMPDIR/wiki-stop-capture-<session_id>.done`
- If the sentinel exists → exits 0 (already nudged this session, no-op)
- On first trigger → creates the sentinel, then emits the nudge (exit 2)
- Falls back gracefully when `session_id` is absent (no sentinel, old behavior)

This is Option B from #91. Chosen over Option A (checkpoint mtime) because it's fully self-contained in the hook — no cross-file dependency that can break on future skill renames (as happened with #82).

## Test plan

- [ ] New session → make ≥1 edit → Stop → nudge fires once ✓
- [ ] Continue session → make more edits → Stop again → silent (sentinel present) ✓
- [ ] New session (new session_id) → sentinel is absent → nudge fires again ✓
- [ ] Tested on st3ve remote sandbox

Closes #91
Refs #82, #81, #85

### DIFF
--- a/.claude/hooks/wiki-stop-capture.sh
+++ b/.claude/hooks/wiki-stop-capture.sh
@@ -23,6 +23,16 @@ print('1' if d.get('stop_hook_active') else '0')
 " 2>/dev/null || echo "0")
 [[ "$IS_HOOK_TURN" == "1" ]] && exit 0
 
+# Fire at most once per session — sentinel file keyed to session_id prevents
+# repeated nudges after the threshold is crossed on the first turn.
+SESSION_ID=$(printf '%s' "$INPUT" | python3 -c "
+import json, sys; print(json.load(sys.stdin).get('session_id', ''))" 2>/dev/null || echo "")
+SENTINEL=""
+if [[ -n "$SESSION_ID" ]]; then
+  SENTINEL="${TMPDIR:-/tmp}/wiki-stop-capture-${SESSION_ID}.done"
+  [[ -f "$SENTINEL" ]] && exit 0
+fi
+
 TRANSCRIPT_PATH=$(printf '%s' "$INPUT" | python3 -c "
 import json, sys
 d = json.load(sys.stdin)
@@ -70,6 +80,7 @@ BASH_COUNT=$(echo "$COUNTS" | awk '{print $2}')
 # Trigger if any file was written/edited, or if there were ≥ 4 shell calls
 # (suggesting investigation/debugging worth preserving).
 if [[ "${WRITE_EDIT:-0}" -ge 1 ]] || [[ "${BASH_COUNT:-0}" -ge 4 ]]; then
+  [[ -n "$SENTINEL" ]] && : > "$SENTINEL" 2>/dev/null || true
   echo "Session ended with ${WRITE_EDIT} file edit(s) and ${BASH_COUNT} shell call(s). Please run /wiki-capture --quick now to preserve any reusable findings before this context closes." >&2
   exit 2
 fi


Ar9av:
Verified on st3ve:

- No `session_id` in payload: exit 2, nudge fires — fallback intact
- `stop_hook_active` set: exit 0, silent
- Sentinel cycle: fire 1 exits 2, sentinel created; fire 2 same session exits 0, no output

Issue #35, opened 2026-05-08
daily-update.sh silently misses relative source paths when run from outside the vault


Jason-Wheeler:
# `daily-update.sh` silently misses relative source paths when run from outside the vault

## Summary

`scripts/daily-update.sh` checks source freshness by calling `os.path.exists(path)` against keys from `.manifest.json`. When the manifest stores a source path as relative to the vault root (the common case), the check only resolves correctly if the current working directory is inside the vault. Run from anywhere else — including the launchd agent the v2026.05 setup mode installs — and every relative path is silently skipped, so `stale_count` stays at `0` even when sources have new content.

This affects the cron path materially: launchd's default `WorkingDirectory` is `/`, so the cron the skill installs reports "Wiki is up to date" forever.

## Reproduction

```bash
# from inside a vault with relative paths in .manifest.json
cd "$OBSIDIAN_VAULT_PATH"
bash scripts/daily-update.sh
# → [wiki-daily] N source(s) have new content since last ingest

# same machine, same vault, run from elsewhere
cd /
bash scripts/daily-update.sh
# → [wiki-daily] Wiki is up to date.   ← incorrect when N > 0
```

In our case, 559 of 680 keys in `.manifest.json` were relative paths. From `/`, all 559 silently skip the existence check, so the freshness count is computed against only the 121 absolute keys.

## Root cause

`scripts/daily-update.sh` lines ~89–94:

```python
for path, meta in manifest.get("sources", {}).items():
    expanded = os.path.expanduser(path)
    if os.path.exists(expanded):
        mtime = datetime.fromtimestamp(os.path.getmtime(expanded), tz=timezone.utc)
        if mtime > last_updated:
            stale += 1
```

`os.path.expanduser()` only handles `~`. There's no resolution against `OBSIDIAN_VAULT_PATH`, so a relative key like `_raw/foo.md` is checked against `$PWD/_raw/foo.md`. When `$PWD != OBSIDIAN_VAULT_PATH`, that file doesn't exist and the source is silently dropped from the count.

## Suggested fix

Two viable patches; either resolves the issue:

**Option A — `cd` once at the top of the script (one line):**

```diff
 source "$CONFIG_FILE"

 if [[ -z "${OBSIDIAN_VAULT_PATH:-}" ]]; then
   echo "[wiki-daily] OBSIDIAN_VAULT_PATH not set in $CONFIG_FILE — skipping" >&2
   exit 1
 fi

[...]

PR #102, opened 2026-06-14
fix(daily-update): normalize naive last_updated to UTC before stale comparison

### LINKED ISSUE #101: daily-update.sh crashes on timezone-naive manifest last_updated (offset-naive vs offset-aware datetime)
## Summary

The source-freshness check in `scripts/daily-update.sh` crashes with `TypeError: can't compare offset-naive and offset-aware datetimes` whenever `.manifest.json`'s top-level `last_updated` is timezone-**naive**. This is a sibling of the already-fixed #35 / #36 — a second latent bug in the same `stale_count` Python block.

## Where

`scripts/daily-update.sh`, the embedded Python that computes `stale_count`:

```python
last_updated_str = manifest.get("last_updated", "")
try:
    last_updated = datetime.fromisoformat(last_updated_str.replace("Z", "+00:00"))
except Exception:
    print(0)
    sys.exit()
...
    mtime = datetime.fromtimestamp(os.path.getmtime(expanded), tz=timezone.utc)
    if mtime > last_updated:        # <-- TypeError if last_updated is naive
        stale += 1
```

`mtime` is always tz-aware (`tz=timezone.utc`). `last_updated` is only tz-aware when the stored string carries an offset. The `try/except` guards the *parse*, not the *comparison*, so a naive value sails through `fromisoformat` and then throws at `mtime > last_updated`, aborting the run.

## Why it triggers / why it's been latent

The documented convention is Z-suffixed (e.g. `wiki-status/SKILL.md` shows `"last_updated": "2026-04-06T10:30:00Z"`), and `.replace("Z", "+00:00")` makes that aware — so the happy path never hits it. But any writer that emits a naive timestamp exposes it. A common way to produce one is `datetime.now().isoformat()` (no offset) when an ingest/maintenance step refreshes `last_updated`. The result is a hard crash of the daily freshness check (and the launchd/systemd job that runs it).

## Reproduce

1. In a vault's `.manifest.json`, set `"last_updated": "2026-06-13T04:27:28"` (no `Z`/offset).
2. Ensure at least one registered source exists on disk.
3. Run `scripts/daily-update.sh` from the vault dir.

Result: `TypeError: can't compare offset-naive and offset-aware datetimes`.

## Fix

Normalize a naive `last_updated` to UTC right after parsing, so the comparison is always aware-vs-aware:

```python
last_updated = datetime.fromisoformat(last_updated_str.replace("Z", "+00:00"))
...
if last_updated.tzinfo is None:
    last_updated = last_updated.replace(tzinfo=timezone.utc)
```

One line, no behavior change for already-aware timestamps. Verified locally: with the guard, a naive `last_updated` runs clean and returns a correct stale count instead of crashing.

## Related

- #35 / #36 — relative source paths silently skipped in the same `stale_count` block (different failure mode, already merged as `aa264a4`).

Ar9av:
## What broke

`stale_count` in `scripts/daily-update.sh` crashes with `TypeError: can't compare offset-naive and offset-aware datetimes` when `.manifest.json`'s `last_updated` has no timezone offset. `datetime.now().isoformat()` produces this format: `2026-06-13T04:27:28`.

`mtime` is always UTC-aware. `last_updated` is only aware when the stored string carries an offset. The `try/except` guards the parse, not the comparison, so a naive value reaches `mtime > last_updated` and throws, killing the daily run and any launchd/systemd job that wraps it.

## Fix

One line after `fromisoformat`:

```python
if last_updated.tzinfo is None:
    last_updated = last_updated.replace(tzinfo=timezone.utc)
```

Already-aware timestamps pass through unchanged. Naive ones get assumed UTC, matching `mtime`'s reference frame.

## Reproduce

```bash
# Set last_updated in .manifest.json to a naive timestamp
# "last_updated": "2026-06-13T04:27:28"
# Then run:
scripts/daily-update.sh
# Before: TypeError crash
# After: correct stale count
```

Closes #101. Sibling of #35/#36 (relative path skipping in the same block, merged in `aa264a4`).

### DIFF
--- a/scripts/daily-update.sh
+++ b/scripts/daily-update.sh
@@ -84,6 +84,8 @@ try:
 except Exception:
     print(0)
     sys.exit()
+if last_updated.tzinfo is None:
+    last_updated = last_updated.replace(tzinfo=timezone.utc)
 
 vault_path = os.environ.get("OBSIDIAN_VAULT_PATH", "")
 stale = 0


cnndabbler (approved):
LGTM — exactly the fix from the issue, in the right place (normalize right after the parse, before the comparison) and `timezone` is already imported. Verified locally: a naive `last_updated` now runs clean instead of throwing. Thanks for the quick turnaround.

PR #116, opened 2026-06-27
fix(pi-history): require session-grounded provenance

yulin0629:
## Problem

`pi-history-ingest` had four structural gaps that let model hallucinations slip through as `^[extracted]` (highest-confidence) claims:

| Gap | Effect |
|---|---|
| No knowledge-closure red line | Model could "helpfully" add package/training/other-session background and present it as session knowledge |
| Provenance defined but never verified | `^[extracted]` became a self-declaration with no external constraint |
| No claim-level source citation | Hallucinations couldn't be traced back to a session entry after the fact |
| Distillation biased toward abstraction | "Merge recurring patterns" / "distill knowledge, not chronology" washed out session-specific decision context |

The skill trusted the model's honesty as the provenance mechanism. Any system that relies on self-marking without verification fails this way, regardless of model intelligence.

## Fix

- **Session knowledge closure** rule: Pi session files are the only factual source; outside context must be an open question or skipped, never presented as extracted.
- **Evidence ledger**: each durable fact carries `pi:<session-file>#<entry-id>` (or `:line<N>`) before any page is written.
- **Source reference comments** on extracted claims: `<!-- source: pi:... -->` as an audit trail.
- **Source verification gate** before writing: every extracted claim has a reference; proper nouns/commands/filenames in extracted claims must appear in the cited entry (literal `grep`/`rg` when in doubt); unverifiable claims get deleted or downgraded to `^[inferred]`.

## Scope

Only `.skills/pi-history-ingest/SKILL.md`. No code, no other skills touched. The same closure gap likely exists in the sibling `*-history-ingest` skills (claude/codex/hermes/openclaw/copilot) — happy to port the same fix there in follow-ups if this approach is accepted.

Verified: `uv run --with pytest pytest -q` → 3 passed.

### DIFF
--- a/.skills/pi-history-ingest/SKILL.md
+++ b/.skills/pi-history-ingest/SKILL.md
@@ -12,6 +12,8 @@ description: >
 
 You are extracting knowledge from the user's Pi coding agent sessions and distilling it into the Obsidian wiki. Pi sessions are stored as structured JSONL with a tree layout — your job is to follow the active branch, extract durable knowledge, and compile it.
 
+**Session knowledge closure:** Pi session files are the only factual source for this skill. Do not add background knowledge from model training, other tools, package docs, local files, or the current conversation unless that fact appears in the selected session entries. If outside context seems useful, mark it as an open question or skip it — never present it as extracted session knowledge.
+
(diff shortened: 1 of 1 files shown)


Ar9av:
LGTM 👍

PR #129, opened 2026-07-09
Prefer ripgrep in wiki retrieval guidance

mike840609:
## Summary
- Add shared retrieval guidance to prefer `rg` / `rg --files` for shell and file searches when available.
- Document `grep` / `find` as the fallback path.
- Clarify that capitalized `Grep` / `Glob` references remain agent-tool-generic primitives.

## Why
- `rg` is generally faster than GNU grep for recursive codebase search; ripgrep's benchmarks show materially better search times across large source trees and large files: https://ripgrep.dev/benchmarks/
- Some self-hosted OpenCode-style workflows may follow grep-oriented skill wording literally rather than automatically choosing shell `rg`, so making the preference explicit helps agents take the faster path when it exists.

## Validation
- Checked diff against `upstream/main`.
- Confirmed this PR changes only `.skills/llm-wiki/SKILL.md`.

### DIFF
--- a/.skills/llm-wiki/SKILL.md
+++ b/.skills/llm-wiki/SKILL.md
@@ -460,6 +460,8 @@ Reading the vault is the dominant cost of every read-side skill. Use the cheapes
 | Whole-page content | `Read <file>` | **Expensive** — last resort |
 | Relationships across pages | `Grep "\[\[.*?\]\]"` across the vault, or walk wikilinks from a known page | Case-by-case |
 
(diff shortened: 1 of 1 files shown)


Ar9av:
Tested on st3ve (Ubuntu Lightsail box, ripgrep 14.1.0 installed) against the live 52-page Knowledge vault:

- `rg --files` page enumeration ✅
- wikilink relationship primitive `rg -o "\[\[.*?\]\]"` ✅
- `rg -A/-B` context primitive behaves exactly as the retrieval-primitives table describes ✅
- **rg/grep parity: identical match counts (27 = 27)** — no behavioral drift vs the existing `grep` guidance
- Fallback path is valid: `grep`/`find` both present

Well-scoped docs-only change, correctly placed right after the retrieval table, and the `Grep`/`Glob` tool-generic clarification is a nice touch. Merging. 🚀

PR #138, opened 2026-07-14
feat(skills): snapshot vault before destructive writes

mike840609:
## Summary

- snapshot Git-backed vaults before `cross-linker`, `wiki-dedup`, and `wiki-lint --consolidate` perform destructive writes
- require the vault itself to be the repository root so `git add -A` cannot capture unrelated parent-repository changes
- distinguish a clean repository from a failed snapshot and abort before vault writes when staging or committing fails
- record an actionable snapshot SHA and document complete rollback commands
- add contract tests covering all three skills

## Why

These skills can rewrite many vault pages in one run. A pre-write commit provides a cheap rollback point, but the initial implementation treated any directory inside a Git worktree as a standalone vault repository. From a nested vault, `git add -A` can stage sibling files outside the vault. It also treated an empty commit's non-zero exit ambiguously and did not provide a complete rollback command.

This version scopes the snapshot to standalone vault repositories and fails closed when the promised snapshot cannot be created.

## Validation

- `216 passed, 11 subtests passed`
- shell syntax checked for all three embedded snapshot blocks
- exercised clean, dirty, nested, non-Git, commit-failure, and rollback scenarios for every affected skill

Tracks the fork issue: [mike840609/obsidian-wiki#4](https://github.com/mike840609/obsidian-wiki/issues/4)

Closes mike840609/obsidian-wiki#4

### DIFF
--- a/.skills/cross-linker/SKILL.md
+++ b/.skills/cross-linker/SKILL.md
@@ -97,13 +97,40 @@ Only act on **EXTRACTED** and **INFERRED** candidates. Include the confidence la
 
 ## Step 4: Apply Links
 
-**Pre-write snapshot** — before the first file write, check `git -C "$OBSIDIAN_VAULT_PATH" rev-parse --is-inside-work-tree`. If the vault is not a git repo, skip this step silently — no nagging, no suggesting `git init`. If it is, snapshot it:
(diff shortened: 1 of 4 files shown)


Ar9av:
thanks lgtm

PR #141, opened 2026-07-15
fix(skills): replace hardcoded grep examples with rg

mike840609:
## Summary
- #129 added a "prefer ripgrep" policy to `llm-wiki/SKILL.md`, but six other skill files still hardcoded literal `grep` commands in their instruction examples, silently bypassing that policy.
- Swaps those literal examples to `rg`, with correct flag mapping (`--include="*.md"` → `--glob '*.md'`; `-r` dropped since `rg` recurses by default).
- Files touched: `wiki-agent`, `wiki-update`, `wiki-ingest`, `wiki-lint`, `wiki-synthesize`, `memory-bridge`.
- Left untouched intentionally: `llm-wiki/SKILL.md`'s `grep -q` config-resolution check (zero-dependency by design) and `claude-history-ingest/SKILL.md`'s mention of `grep -v` (an anti-pattern warning, not an instruction).

## Behavior note
`rg` respects `.gitignore` and skips hidden files by default; plain `grep -r` doesn't. This only matters for the two directory-wide backlink scans in `wiki-lint`/`wiki-synthesize` (`wiki-agent`'s single-file search, the piped `qmd ls` filters, and `memory-bridge`'s explicit-file-list search never traverse a directory, so it doesn't apply to them).

**Concrete effect on this vault:** the standard vault `.gitignore` includes `.trash/`. Before this change, `grep -r` would count wikilink backlinks from pages sitting in Obsidian's trash toward co-occurrence scores; after this change, `rg` silently excludes them. That's a real, not just hypothetical, behavior difference for anyone with deleted pages still in `.trash/` — it's arguably the more correct behavior for synthesis-gap detection (deleted pages shouldn't count), but it is a change, and reviewers should decide if that's the intended semantics. Aside from `.trash/`, a typical vault's `.gitignore` only covers Obsidian internals (`.obsidian/workspace.json`, etc.), not content pages, so the two commands scan the same `.md` file set otherwise.

## Test plan
- [x] Ran the old `grep -rl ... --include="*.md"` vs new `rg -l --glob '*.md'` against real vault data — identical file match sets
- [x] Ran `grep -i` vs `rg -i` on a real session `.jsonl` file — byte-identical output
- [x] Ran the piped `cmd | grep "substr"` vs `cmd | rg "substr"` pattern — identical filtered output

### DIFF
--- a/.skills/memory-bridge/SKILL.md
+++ b/.skills/memory-bridge/SKILL.md
@@ -87,7 +87,7 @@ Read frontmatter for the listed pages (grep for `^(title|category|tags|updated):
 
 Within the filtered page set, run:
 ```
-grep -l "<topic>" <pages in tool set>
+rg -l "<topic>" <pages in tool set>
 ```
 Then grep section headers (`^##`) around matches to give context without full reads. Present results as a ranked list with the matching excerpt.
 
--- a/.skills/wiki-agent/SKILL.md
+++ b/.skills/wiki-agent/SKILL.md
@@ -131,20 +131,20 @@ Open each selected session file and extract only the content relevant to the que
 
 **Claude** (JSONL conversation):
 - Each line: `{role, content, timestamp, ...}`
-- Search with: `grep -i "<query terms>" <session.jsonl>` to find the relevant lines
+- Search with: `rg -i "<query terms>" <session.jsonl>` to find the relevant lines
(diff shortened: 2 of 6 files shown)


Ar9av:
makes sense , ripgrep is cool

PR #142, opened 2026-07-15
docs: advisory README translation drift check with backfill script

mike840609:
## What changed

- **The README sync check is advisory — it never blocks PR merges.** The `readme-translation-drift (advisory)` job surfaces drift as a `::warning` annotation only and always succeeds.
- `tools/check_readme_sync.py` is now a drift reporter: it lists the commits that changed `README.md` after the last commit that touched `README_TW.md`, plus the combined English diff still pending translation. Run it locally, translate the diff, and backfill `README_TW.md`.
- A later translation-only commit to `README_TW.md` clears the drift, so translations can land asynchronously instead of being forced into the same PR as the English change.
- Documented the advisory workflow in `AGENTS.md` and both README contributor sections.
- `tests/test_readme_sync.py` exercises the drift semantics against a temporary git repo: in sync, English-only drift reported with diff, catch-up translation commit clears drift, unrelated commits ignored, and the workflow is documented in all three docs.

## Why non-blocking

A hard "both files must change together" gate forces every docs PR to bundle a translation, which blocks contributors who can't write Traditional Chinese. The advisory model keeps the parity goal visible — a recurring CI warning on every PR/push until the translation catches up, plus an on-demand local report — while letting translation land as a follow-up.

## Impact

No PR is ever failed by this check. While `README_TW.md` is behind, every CI run (any PR, any push to `main`) re-reports the warning until a commit touches `README_TW.md`.

Known limitation (marked `ponytail:` in the script): a Chinese-only edit (e.g. a typo fix) also counts as catching up and clears earlier drift; upgrade to per-commit pairing if that ever misleads.

## Validation

- `uv run --no-project --with pytest --python 3.12 pytest -q` — 218 passed, 14 subtests passed
- `python3 -m unittest tests.test_readme_sync -v` — 5 passed
- `python3 tools/check_readme_sync.py` on this branch — in sync, exit 0

## Follow-up

None required — the check is deliberately not part of any required ruleset. Do **not** add `README sync / readme-translation-drift` as a required status check; that would defeat the advisory design.

### DIFF
--- a/.github/workflows/readme-sync.yml
+++ b/.github/workflows/readme-sync.yml
@@ -29,11 +29,14 @@ jobs:
 
       - name: Require README translations to change together
         env:
+          EVENT_NAME: ${{ github.event_name }}
           BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
         run: |
-          if [[ "$BASE_SHA" =~ ^0+$ ]]; then
-            git diff-tree --root --no-commit-id --name-only -r "$HEAD_SHA"
+          if [[ "$EVENT_NAME" == "pull_request" ]]; then
+            git diff --name-only "$BASE_SHA...$HEAD_SHA"
+          elif [[ "$BASE_SHA" =~ ^0+$ ]]; then
+            git ls-tree -r --name-only "$HEAD_SHA"
           else
             git diff --name-only "$BASE_SHA" "$HEAD_SHA"
           fi | python tools/check_readme_sync.py
--- a/.github/workflows/readme-sync.yml
+++ b/.github/workflows/readme-sync.yml
@@ -1,0 +1,39 @@
+name: README sync
+
+on:
+  pull_request:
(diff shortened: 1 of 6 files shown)


mike840609:
~Hi @Ar9av,  unsure if this change will slow down other developers; should it be made non-mandatory?
I can periodically check file consistency by script. LoL~

Ar9av:
You'\''re right to have raised this, and the latest commit (`e817220`, 7/18) already addresses it — the checker is now advisory-only, not a merge gate.

Verified locally:
- `readme-sync.yml`'\''s content-check step is `python tools/check_readme_sync.py || echo "::warning ..."` — the `echo` fallback always exits 0, so the job can'\''t fail from README drift regardless of the script'\''s exit code.
- `tests/test_readme_sync.py` — 5/5 pass.
- `tools/check_readme_sync.py` run directly — correctly reports drift/up-to-date.
- `main` currently has no branch protection / required status checks configured, so this couldn'\''t block a merge even if it weren'\''t advisory by design.

No further changes needed — this looks good to merge as-is.

PR #161, opened 2026-08-04
fix(hooks): harden the read-only classifier; re-arm the nudge for long sessions

selivanov-tech:
Follow-up to #158, two commits:

1. **19 read-only-classifier escapes closed** — this commit was pushed to the #158 branch about an hour after the merge, so it missed the train.
2. **Re-arm for long-running sessions** — fixes the starvation the once-per-session sentinel introduced.

## Commit 1 — classifier escapes

An adversarial review pass hunted for commands and tool calls the read-only classifier would misclassify as safe, and found 19 concrete escapes. All are closed, each pinned by a transcript-level test.

**Tool-call layer**
- **Task/Agent are no longer blanket-neutral** — a spawned subagent can edit files or call write APIs. The classifier now reads `subagent_type`: only read-only agent types (Explore, Plan) keep the exemption.
- **EnterWorktree/ExitWorktree** removed from the neutral set (they create/remove branches and worktrees).
- **Read-verb SQL tools scan their payload**: `EXPLAIN ANALYZE` executes the statement, so DML keywords or `"analyze": true` in the input disable the exemption; plain `SELECT` explains stay neutral.
- Verb lists: `prepare` moved to write verbs (`prepare_query_tuning` provisions a branch); `resolve` dropped from read verbs (resolving a review thread mutates it).

**Quote scanner**
- Double-quote escaping now uses **backslash parity**: `"a\\"` closes the quote (the backslashes escape each other), so a command hidden after `;` is seen. Previously any backslash before the quote kept the scanner in-quote and swallowed the rest of the command.

**Command classifiers**
- **python -c**: `.execute()/.executemany()`, `Path.open('w')`, aliased os imports (`import os as x`, `from os import …`); matching is now case-insensitive so lowercase `delete from` in a sqlite call is caught.
- **curl**: `--config`/`-K` (a config file can carry any request), `-Q`/`--quote` (FTP protocol commands like `DELE`), with Q/K detected inside short-flag clusters.
- **git**: `--upload-pack`/`--receive-pack`/`--exec` inject remotely-executed commands even under read-only subcommands (`git ls-remote --upload-pack='…'`).
- **date**: a positional operand sets the clock on BSD/macOS (`sudo date 010100002025`); `+format`, `-u`, `-j` forms stay read-only.
- **awk**: redirects/pipes to *variables* (`print > f`, `print | c`), `close()`, `getline` — numeric comparisons (`$3 > 100`) stay clean.
- **sed**: the `w` script command is caught with any substitution delimiter (`s#a#b#w out`), not just `/`.
- **xxd/uniq**: a bare `-` stdin marker now counts as a positional, so the *output file* argument is no longer missed (`xxd -r -p - out.bin`).
- **less/more**: `-o`/`-O`/`--log-file` copy input into a file.

## Commit 2 — re-arm for long sessions

"Once per session" starves multi-day sessions: the single nudge fires early (usually after the first few edits), and all later work is never captured automatically.

The sentinel now **expires** when BOTH hold:

- it is older than `WIKI_STOP_REARM_SECONDS` (default **6h**), and
- at least `WIKI_STOP_REARM_EDITS` (default **10**) new file edits happened since the nudge that claimed it.

| Session shape | Behavior |
|---|---|
| Short session | one nudge, exactly as before |
| Long, busy session | roughly one nudge per work stretch |
| Long, idle session | one nudge, then silence |

Mechanics: the edit count at nudge time is stored inside the sentinel directory; the young-sentinel path stays the cheap pre-parse short-circuit; an expired sentinel is retired with an atomic `mv`, so duplicate hook registrations still produce exactly one re-nudge. Sentinels predating the feature have no stored count and re-arm from zero. Both thresholds are env-overridable.

## Tests

Suite grows to **73** (20 for the classifier escapes, 6 for re-arm — age gate, delta gate, count storage and reset, pre-feature sentinels, env knobs). All passing; hook stays bash-3.2 compatible.

Design rule unchanged: every classifier fix errs toward *mutating* — unknown or ambiguous means the nudge fires.

### DIFF
--- a/.claude/hooks/wiki-stop-capture.sh
+++ b/.claude/hooks/wiki-stop-capture.sh
@@ -108,32 +108,48 @@ GH_READONLY = {"view", "list", "status", "diff", "checks"}
 # notably MCP tools without a clear read verb — makes the session
 # ineligible for the read-only exemption, because a real system mutation
 # (a CRM update, an SQL INSERT) may hide behind it.
+# Task/Agent are NOT here: a spawned subagent can edit files or call write
+# APIs, so they get argument-aware handling in the counting loop (only the
+# read-only agent types Explore/Plan stay neutral). EnterWorktree/ExitWorktree
+# are not here either — they create/remove git branches and worktrees.
 CORE_READONLY_TOOLS = {
     "Read", "Glob", "Grep", "LS", "WebFetch", "WebSearch", "TodoWrite",
-    "TodoRead", "NotebookRead", "Task", "Agent", "AskUserQuestion",
+    "TodoRead", "NotebookRead", "AskUserQuestion",
     "ToolSearch", "Skill", "SkillSearch", "SlashCommand", "EnterPlanMode",
-    "ExitPlanMode", "EnterWorktree", "ExitWorktree", "Explore", "Plan",
+    "ExitPlanMode", "Explore", "Plan",
     "TaskCreate", "TaskUpdate", "TaskGet", "TaskList", "TaskOutput",
     "TaskStop", "BashOutput", "KillShell", "SendUserFile", "Monitor",
     "ScheduleWakeup", "ListMcpResourcesTool", "ReadMcpResourceTool",
     "ReadMcpResourceDirTool",
 }
+READONLY_AGENT_TYPES = {"Explore", "Plan"}
+# "resolve" is deliberately absent from the read verbs: resolving a review
+# thread mutates it, so resolve-named tools stay suspicious.
 MCP_READ_VERBS = {
     "get", "list", "search", "read", "query", "fetch", "find", "describe",
     "inspect", "explain", "compare", "view", "check", "status", "whoami",
-    "resolve", "analyze", "show", "health", "info", "download", "lookup",
+    "analyze", "show", "health", "info", "download", "lookup",
     "browse", "watch", "screenshot",
 }
 # Write verbs take priority over read verbs for names carrying both
-# ("get-or-create-session" creates).
+# ("get-or-create-session" creates; "prepare_query_tuning" provisions).
 MCP_WRITE_VERBS = {
     "create", "update", "delete", "remove", "write", "insert", "upsert",
     "execute", "run", "send", "post", "put", "patch", "move", "add", "set",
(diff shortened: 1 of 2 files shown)


Ar9av:
4 of the re-arm tests fail on Linux:

- `test_rearm_after_age_and_new_edits`
- `test_rearm_updates_stored_count_and_rests_again`
- `test_pre_feature_sentinel_rearms_from_zero`
- `test_concurrent_rearm_produces_exactly_one_nudge`

Root cause: the `stat -f %m FILE || stat -c %Y FILE` fallback, used twice for `CLAIMED_AT` and `RET_AT`. GNU `stat`'s `-f` flag doesn't take a format argument; it means "show filesystem status." On Linux, `%m` gets parsed as a second FILE operand. That operand errors, but the command still prints real filesystem info for the sentinel to stdout and exits nonzero overall. So `||` runs `stat -c %Y` too, and `$(...)` concatenates both outputs into one non-numeric string. The `^[0-9]+$` guard rejects that string and falls back to `$NOW`. Every sentinel then looks freshly-created, and the age check never passes. The re-arm feature never fires on Linux.

Reproduced by hand:
```
$ stat -f %m wiki-stop-capture-dbg.done 2>/dev/null || stat -c %Y wiki-stop-capture-dbg.done
  File: "wiki-stop-capture-dbg.done"
    ID: df5af021296f7270 Namelen: 255     Type: ext2/ext3
...
1785837805
```

Pushed a fix (`8de30c7`) that flips the order: try `stat -c %Y` first, fall back to `stat -f %m`. GNU-style `-c` is an invalid option on BSD/macOS `stat`, so it fails clean there too: nonzero exit, no stdout. The reverse order only fails clean on the platform whose flag came first. All 74 tests now pass on both macOS and Linux.

One more thing, not blocking: the broadened `AWK_RISKY` regex (`>` / `|` followed by `["'A-Za-z_]`) now also matches plain variable-to-variable awk comparisons like `a > b`, not just redirects to variables. `NR > 1` and `$3 > 100` stay clean because a literal digit follows the operator, but `$3 > head` flags as mutating since `head` is a variable name. The PR's stated design rule is "ambiguous means the nudge fires," so this fits the intent, but scripts that compare across variables will nag more than expected.

PR #164, opened 2026-08-08
Stabilize Windows vault setup

kidongnam1:
## Summary
- Preserve existing obsidian-wiki config values across setup reruns while still updating the managed vault/repo/version keys.
- Honor HOME/USERPROFILE in portable and test contexts on Windows.
- Stabilize the Claude stop hook when `python3` is unavailable and make stop-hook behavior tests use Git Bash-compatible paths on Windows.
- Normalize manifest relative-key matching across Windows and POSIX separators.
- Document the default external document-library mode: link/index originals by path and hash instead of moving source files.
- Add LF line-ending policy for shell, Python, Markdown, config, and JSON files.

## Why
Windows users can have WSL Bash, Git Bash, and Python launchers on the same machine. The previous setup/test assumptions could miss the intended config, silently skip stop-hook capture, or misclassify relative manifest keys. Existing business document folders also need a non-destructive default ingest policy.

## Validation
- `python -m pytest -q` -> 520 passed, 19 subtests passed
- `python -m obsidian_wiki.cli doctor --json` -> pass
- `git diff --check` -> clean
- `git ls-files --eol` verified LF for touched shell/Python/Markdown files

## Notes
This PR leaves source document libraries in place by default. `stage-copy` is documented as an explicit review-copy mode, while moves require separate user approval and backup/hash verification.

### DIFF
--- a/.claude/hooks/wiki-stop-capture.sh
+++ b/.claude/hooks/wiki-stop-capture.sh
@@ -22,10 +22,18 @@
 
 set -euo pipefail
 
+if command -v python3 >/dev/null 2>&1; then
+  PYTHON_BIN="python3"
+elif command -v python >/dev/null 2>&1; then
+  PYTHON_BIN="python"
+else
+  exit 0
+fi
+
 INPUT=$(cat)
 
 # Suppress if already in a stop-hook-triggered turn (prevents infinite loops)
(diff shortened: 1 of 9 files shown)


Ar9av:
keep going for it, I dont have a windows OS to test, but can you update and check thoroughly before the PR

Ar9av:
@kidongnam1 let me know if you are picking this up?

Issue #168, opened 2026-08-11
Improve /wiki-update coverage for large projects


mike840609:
## Problem

`/wiki-update` can miss entire components when building knowledge for a large project, even on a first-time "full scan".

The skill currently asks the agent to scan the project heuristically (README/docs/source structure/package metadata/git history) and then distill only high-value knowledge. That works well for smaller repositories, but there is no deterministic coverage check that verifies every major component/subsystem was at least inspected.

This creates two related failure modes:

1. A component may be skipped entirely because the agent focuses on representative or high-signal areas.
2. After the first run, `.manifest.json` records `last_commit_synced = HEAD`. If an omitted component has no later commits, future delta-based `/wiki-update` runs may never revisit it.

This means an incomplete first scan can become persistent.

## Current behavior

The current flow is roughly:

```text
README / docs / source structure
        ↓
heuristic exploration
        ↓
select high-value knowledge
        ↓
distill pages
        ↓
record project-level last_commit_synced
```

There is no explicit record of which components were inspected, distilled, or intentionally skipped.

## Proposed improvement

Add a lightweight **component-aware coverage pass** before distillation, while keeping the existing Karpathy-style principle of distilling knowledge rather than copying implementation details.

Suggested flow:

```text
Structural census
  (git ls-files / rg --files)
        ↓
Build component/subsystem map
        ↓
Inspect each major component
        ↓
Distill only meaningful knowledge
        ↓
Coverage check
        ↓
Update last_commit_synced
```

The goal is not to read every line or create a wiki page for every folder. The important distinction is:

- `inspected-no-distill`: the component was considered and contained no knowledge worth persisting
- `distilled`: useful knowledge was captured
- `pending`: the component was discovered but not yet inspected

## Possible manifest extension

A minimal project-level structure could be:

```json
{
  "components": {
    "auth": {
      "paths": ["src/auth"],
      "status": "distilled"
    },
    "reporting": {
      "paths": ["src/reporting"],
      "status": "inspected-no-distill"
    },
    "scheduler": {
      "paths": ["src/jobs"],
      "status": "pending"
    }
  }
}
```

This does not need to track every source file individually.

## Acceptance criteria

- First-time `/wiki-update` discovers the repository's major components/subsystems before distillation.
- Every discovered component is classified as at least `distilled`, `inspected-no-distill`, or `pending`.
- `last_commit_synced` is not treated as proof of full repository coverage when pending components remain.
[...]

PR #169, opened 2026-08-14
feat: improve project ingest with optional code-graph-assisted understanding (#167)

### LINKED ISSUE #167: feat: Improve project ingest with optional code-graph-assisted understanding
## Summary

Improve `wiki-update` / project ingest for medium and large codebases by optionally using a code graph backend to understand project structure, dependencies, call paths, and affected components before distilling knowledge into the wiki.

The goal is **not to persist the raw source-code graph into the vault**. The graph is an implementation detail used to help the agent decide what code is important to inspect and understand.

## Why

Today, `wiki-update` understands a project mainly through:

- README / docs
- repository structure
- manifests
- git history
- targeted source reads / grep

This works well for small projects, but on larger repositories it can miss important components or relationships because the agent has to discover the architecture while deciding which files to read.

For example, a change in:

```text
src/auth/token.py
```

may affect:

```text
AuthMiddleware
    ↓
AuthService
    ↓
TokenValidator
    ↓
UserRepository
```

A plain file or text search may find some of these pieces, but it does not naturally expose the dependency/call topology.

A code graph can provide that structural context before the LLM starts reading implementation details.

## Proposed behavior

Add an optional **code-understanding backend** to project ingest.

```text
Codebase
   ↓
Code understanding backend
   ↓
Relevant components / call paths / dependencies
   ↓
Selective source reads
   ↓
LLM distillation
   ↓
Obsidian Wiki
```

### Default behavior

Existing behavior remains fully supported:

```text
git diff
+ built-in ast-extract
+ rg / grep
+ selective reads
```

No additional dependency is required.

### Enhanced behavior

When a supported code graph backend is available, project ingest may use it to:

- identify important modules/components
- trace callers and callees
- inspect dependency neighborhoods
- understand cross-file relationships
- determine which files are worth reading
- expand changed files into their likely impact area

For example:

```text
git diff
   ↓
changed symbols/files
   ↓
code graph impact / neighborhood
   ↓
rank relevant components
   ↓
read selected code
   ↓
distill durable knowledge
```

The raw graph should **not** be copied into the vault by default.

The vault continues to store human-oriented, durable knowledge rather than implementation-level symbol graphs.

## Backend

For the first implementation, support an optional CodeGraph-style backend such as `@colbymchenry/codegraph`.

The integration should be provider-oriented rather than tightly coupling `wiki-update` to one implementation:

```text
code understanding
       │
       ├── enhanced backend: CodeGraph
       │
       └── built-in fallback: ast-extract + rg
```

Suggested configuration:

```text
CODE_UNDERSTANDING_BACKEND=auto
```

Behavior:

- `auto` — use CodeGraph when available, otherwise built-in
- `builtin` — always use `ast-extract` + `rg`
- `codegraph` — explicitly require CodeGraph; warn if unavailable

## Incremental behavior

A persistent local code index should be treated as a **cache / sidecar**, not wiki knowledge.

Example:

```text
project/
├── .git/
├── .codegraph/
└── src/
```

`wiki-update` should continue using the existing project sync commit/delta information to avoid unnecessary work.

On subsequent updates:

```text
last synced commit
      ↓
git diff
      ↓
changed files/symbols
      ↓
refresh/query affected graph region
      ↓
select impacted components
      ↓
update relevant wiki knowledge
```

The intent is to avoid re-discovering the entire codebase on every update.

## Context selection

Do not dump the entire graph into the LLM context.

Use an Aider repo-map-style approach:

1. Start from changed files, user intent, or important entry points.
2. Query nearby dependencies/callers/callees.
3. Rank important components.
4. Read only the most relevant implementation.
5. Distill knowledge from those reads.

This keeps graph-assisted ingest useful on large repositories without replacing one context-size problem with another.

## Citations / evidence

When structural relationships influence generated wiki knowledge, the final knowledge should still be grounded in source evidence.

Code graph results should guide **where to look**, but important architectural claims should be backed by source files, docs, configuration, or other inspectable evidence.

If relationships are later persisted into the wiki graph, every persisted relationship should carry a citation/evidence reference, and relationships that are no longer supported by the current codebase must be removable during later syncs.

## Doctor integration

Extend:

```bash
obsidian-wiki doctor --project .
```

with a code-understanding capability section.

Example with CodeGraph:

```text
Code understanding
  ✓ Built-in AST extractor
  ✓ rg
  ✓ CodeGraph available
  ✓ Project index initialized
  ✓ Index up to date

Mode: enhanced (CodeGraph)
```

Without CodeGraph:

```text
Code understanding
  ✓ Built-in AST extractor
  ✓ rg
  ○ CodeGraph not installed (optional)

Mode: basic
```

Missing CodeGraph must **not** make the default doctor check unhealthy.

If the user explicitly selects `CODE_UNDERSTANDING_BACKEND=codegraph`, an unavailable or broken backend should produce a warning/error.

## Non-goals for v1

- Persisting every function/class/import/call edge into the Obsidian vault
- Requiring a graph database
- Replacing Markdown as the source of truth for compiled knowledge
- Requiring CodeGraph for normal `wiki-update`
- Building compiler-grade semantic analysis inside obsidian-wiki
- Dumping the full graph into LLM context

## Future extension

Once graph-assisted project understanding is proven useful, a later feature can selectively persist **architecture-level relationships** such as:

```text
auth-service ──uses──────▶ token-store
api-layer    ──depends_on▶ auth-service
```

These should be distilled relationships between meaningful components, not raw symbol-level edges.

That future representation could then be consumed by `wiki-query`, `wiki-synthesize`, and `wiki-export`.

## Acceptance criteria

- [ ] Existing `wiki-update` behavior remains backward compatible.
- [ ] Project ingest exposes a generic code-understanding backend abstraction.
- [ ] Built-in `ast-extract + rg` remains the dependency-free fallback.
- [ ] An optional CodeGraph backend can be used when available.
- [ ] CodeGraph results guide selective source reads rather than being dumped wholesale into context.
- [ ] Incremental project updates do not require rebuilding understanding of the full repository when only a small area changed.
- [ ] Raw source-code graph data is not persisted into the vault by default.
- [ ] Architectural claims written to the wiki remain grounded in inspectable source evidence.
- [ ] `obsidian-wiki doctor --project .` reports basic vs enhanced code-understanding capability.
- [ ] Missing optional CodeGraph support does not make normal doctor checks fail.
- [ ] Tests cover both enhanced-backend and fallback behavior.

mike840609:
Closes #167

## Summary

Adds an optional **code-graph-assisted understanding** step to `wiki-update` project ingest. Before distilling knowledge into the wiki, the agent can now query a code graph backend to identify important modules, call paths, and impact areas — then read only the load-bearing code, with `file:line` citations on every architectural claim.

**Why it matters:** agents otherwise explore large projects heuristically, reading tens to hundreds of thousands of tokens just to understand structure — and can miss entire components. Measured on two real projects:

| Scenario | [Project 1](https://github.com/mike840609/iiwi)  | [Project 2](https://github.com/mike840609/assets_tracker)  |
|---|---|---|
| **builtin** (no codegraph) — agent explores the project heuristically (status quo) | 1,515,799 chars / **~379k tokens** | 2,312,944 chars / **~578k tokens** |
| **codegraph** — ranked focus map from the real call graph | 2,510 chars / **~627 tokens** | 3,136 chars / **~784 tokens** |
| Selective reads: `file:line` ranges the focus map points at | 10 lines | 11 lines |
| **Context saved per incremental `wiki-update`** | **~378k tokens (~99.8%)** | **~577k tokens (~99.86%)** |
| Cost saved per run (~$2.5–3/M input tokens) | **~$0.95–1.13** | **~$1.44–1.73** |
| Single-file scan latency | <5s | **~0.9s** |

**builtin** is the no-codegraph status quo: it also emits a focus map, but its `rg`-based links are too weak to guide selective reads (they can point at the wrong file/line), so the agent still explores heuristically and the savings shown here are realized with **codegraph**. The 2–3 min full-project index is one-time and only when CodeGraph is enabled.

Beyond tokens, the focus map is **structural, not heuristic**: it is seeded from the real call graph, so high-connectivity modules (hubs) surface even when the agent would have skipped them — reducing the risk of permanently missing a component that `last_commit_synced` would never revisit (the motivation behind #168).

The raw graph is never persisted into the vault. The `.codegraph/` index stays a git-ignored sidecar in the project repo.

### Single-file coverage: builtin vs codegraph

Both outputs below were generated for the **same single file** — they show how much of that file's structure each backend captures.

The difference is **precision, not just coverage**: in the builtin output, `adopt_legacy` is attributed to `history.py:18` (it actually lives in `paths.py:15`), and three `rg-reference` rows all point at the same `cli_actions.py:25` line. The codegraph output resolves the real callees (`_json_default`, `_open_for_append`), the real caller (`resolve` in `services/scan.py`), and the real definition site — so the architecture claims written to the wiki are grounded in fact, not keyword guesses.

#### builtin (rg / grep + regex AST extract)
Screenshot 2026-08-15 at 3 45 54 AM

#### codegraph
Screenshot 2026-08-15 at 3 46 46 AM

### New: `obsidian-wiki code-understand` command
```
obsidian-wiki code-understand --project <dir> [--backend auto|builtin|codegraph]
    [--changed <file>...] [--since <sha>] [--max-symbols N] [--pretty]
```
Emits a ranked **focus map** (symbols + `file:line` citations + evidence type) — no source bodies, so it guides selective reads instead of dumping context. Seeds from changed files / `--since <sha>` delta / all tracked files.

### Backend abstraction
- `builtin` (default, zero deps): regex AST extraction (`ast-extractor`) + `rg` cross-file references
- `codegraph` (optional): shells out to the codegraph CLI (`impact`/`callers`/`callees`, `--json`), defensive parsing. Because it walks the real call graph, it also answers **impact questions** — "if this file changes, which components and tests are affected" — so wiki-update knows what to re-verify after a delta

Selected via `CODE_UNDERSTANDING_BACKEND=auto|builtin|codegraph` (default `auto` = use codegraph when available, else builtin).

### wiki-update Step 3b (optional, guarded)
- Runs `code-understand --since <last_commit_synced>` between deciding what to distill and distilling
- **GUARD**: command unavailable/fails → skip and continue, normal sync unaffected
- Requires `(file:lines)` citations for architectural claims
- Mandatory **pruning rule**: stale relationships (target symbol gone / unreachable) are removed and logged, so false positives don't accumulate (owner requirement)

### doctor integration
`obsidian-wiki doctor --project .` now reports a code-understanding capability section:
`code-understanding.builtin` / `.rg` / `.codegraph` / `.codegraph-index` / `.codegraph-fresh` (basic vs enhanced mode). Missing codegraph is `warn` (exit 0); only explicit `CODE_UNDERSTANDING_BACKEND=codegraph` with a broken backend is `fail` (exit 1).

### Config & docs
- `CODE_UNDERSTANDING_BACKEND` (default `auto`)
- `CODE_UNDERSTANDING_CODEGRAPH_BIN` (optional path when binary isn't on PATH)
- Documented in `.env.example`, `docs/configuration.md`, `docs/cli.md`, and `llm-wiki` env reference; wiki-update Step 3b can offer to install codegraph for the user (with consent)

## Backward compatibility

No codegraph installed → everything works exactly as before: builtin fallback, doctor `warn` not `fail`, plain `doctor` output unchanged, existing commands untouched, zero new dependencies.

## Test evidence

- 15 new unit tests + 8 new CLI subprocess tests + 6 new doctor tests (fake codegraph CLI fixture — no binary needed in CI)
- Full suite: `552 passed, 19 subtests passed`; 4 failures pre-existing on `main` (env-dependent `test_context_pack_cli` ×2, `test_lint` ×2)
- Verified against real codegraph v1.5.0 (ranked focus map with `codegraph-edge` call-path evidence)

### 📸 Section 0 — Environment: no codegraph on PATH, new build
Screenshot 2026-08-15 at 2 39 16 AM

### 📸 Section 1 — `code-understand` auto-falls-back to builtin (exit 0)
Screenshot 2026-08-15 at 2 41 36 AM

### 📸 Section 2 — `doctor --project .` — codegraph check is `warn` not `fail` (exit 0)
Screenshot 2026-08-15 at 2 42 57 AM

### 📸 Section 3 — plain `doctor` — no code-understanding section (unchanged)
Screenshot 2026-08-15 at 2 43 44 AM

### 📸 Section 4 — existing `ast-extract` still works
Screenshot 2026-08-15 at 2 44 13 AM

### 📸 Section 5 — full pytest suite (`552 passed`)
Screenshot 2026-08-15 at 2 45 41 AM

>  4 failures are pre-existing (test_context_pack_cli ×2, test_lint ×2).

### 📸 Section 6 — explicit `CODE_UNDERSTANDING_BACKEND=codegraph` without binary → `fail` (by design)
Screenshot 2026-08-15 at 2 46 37 AM

### 📸 Section 7 — doctor enhanced mode: all five code-understanding checks pass
```
obsidian-wiki doctor --project .
```
Expected:
```
✅ code-understanding.builtin: found 1710 AST node(s)
✅ code-understanding.rg: /opt/homebrew/bin/rg
✅ code-understanding.codegraph: /${PATH_TO_YOUR_CODEGRAPH}/codegraph/bin/codegraph
✅ code-understanding.codegraph-index: fresh
✅ code-understanding.codegraph-fresh: fresh
```
Screenshot 2026-08-15 at 3 16 48 AM

### 📸 Section 8 — codegraph focus map with `codegraph-edge` call-path evidence (real project [iiwi](https://github.com/mike840609/iiwi))
```

obsidian-wiki code-understand --project . --backend codegraph \
  --changed ${EXAMPLE_FILE} --max-symbols 10 --pretty
```
Expected: `backend: codegraph` with `changed-file` seeds first, then `codegraph-edge` entries tracing cross-file callers (`resolve` in `services/scan.py`), callees (`_json_default`, `_open_for_append`), and test dependencies (`_entry` in `tests/unit/test_history.py`).

Screenshot 2026-08-15 at 3 18 03 AM

### 📸 Section 9 — project without an index ([iiwi](https://github.com/mike840609/iiwi)): auto-initialized on first run, incremental after
```
obsidian-wiki code-understand --project . --backend codegraph --changed src/iiwi/history.py --pretty
```
First run auto-runs `codegraph init` (creates `.codegraph/codegraph.db`) and returns the focus map; second run queries incrementally without re-initializing.
Screenshot 2026-08-15 at 3 19 34 AM
> 50 symbols 

### 📸 Section 10 — builtin vs codegraph side-by-side on the same file
```
# builtin (no env var): AST evidence only
obsidian-wiki code-understand --project . --backend builtin --changed ${EXAMPLE_FILE}  --max-symbols 6 --pretty

# codegraph (with env var): adds cross-file edge evidence
obsidian-wiki code-understand --project . --backend codegraph --changed ${EXAMPLE_FILE}  --max-symbols 6 --pretty
```
Screenshot 2026-08-15 at 3 36 03 AM

## Non-goals (unchanged from the issue)
- No raw symbol graph in the vault · no graph database · no compiler-grade analysis · no full-graph context dumps · no MCP integration · no `ast_extractor` call-edge changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -189,6 +189,17 @@ PAGEINDEX_MIN_PAGES=30
 # Optional cache dir for *_structure.json (default: <PAGEINDEX_REPO>/results).
 PAGEINDEX_WORKSPACE=
 
+# --- Code understanding (optional) ---
+#
+# How wiki-update understands project structure before distilling knowledge.
+#   auto:    use CodeGraph when available, else the built-in ast-extract + rg (default)
+#   builtin: always use ast-extract + rg (dependency-free)
+#   codegraph: explicitly require CodeGraph; warn/error if unavailable
+CODE_UNDERSTANDING_BACKEND=auto
+# Optional path to the codegraph binary if it isn't on PATH
+# (e.g. CODE_UNDERSTANDING_CODEGRAPH_BIN=/path/to/codegraph)
+CODE_UNDERSTANDING_CODEGRAPH_BIN=
+
 # --- Raw Staging Directory (optional) ---
 #
 # A directory inside OBSIDIAN_VAULT_PATH for unprocessed draft pages.
--- a/.skills/code-understand/SKILL.md
+++ b/.skills/code-understand/SKILL.md
@@ -1,0 +1,90 @@
+---
+name: code-understand
+description: >
+  Build a ranked, citation-backed focus map of a codebase — the files and symbols an
+  architecture actually hangs on — using CodeGraph when available and a dependency-free
+  builtin (regex AST extraction + `rg` cross-file references) otherwise. Use this any time
+  you (the coding agent) need to orient in an unfamiliar or large codebase before making a
+  change: the user says "help me understand this codebase", "what's load-bearing here",
+  "what would break if I change X", "who calls this", "map out this project's architecture",
+  or you're about to touch code you haven't read yet and want to know what matters before
+  scanning everything by hand. This is the same extractor `wiki-update` Step 3b uses
+  internally — this skill exposes it directly so any agent session can call it on demand,
+  not just during a wiki sync.
+---
+
+# Code Understand — On-Demand Architecture Focus Map
+
+You are about to work in a codebase you don't fully know yet. Instead of grepping around or
+reading files at random, ask the local extractor for a **focus map**: the ranked symbols the
+architecture hangs on, each with a `file:line` citation and evidence type. Read only what it
+points at — this is a map, not a substitute for reading the actual code it cites.
+
+## When to reach for this
+
+- Before editing an unfamiliar module — see what calls it and what it calls before changing
+  its shape.
+- The user asks "how does X work", "what's the impact of changing Y", "who else uses this
+  function" for a codebase, not the wiki.
+- You're about to do a large refactor and want a ranked list of load-bearing files instead of
+  reading the whole tree.
+- You already have a diff or a set of changed files and want to know their blast radius before
+  finishing the change.
+
+Not for: summarizing what a project *does* for the wiki (that's `wiki-update` Step 3b, which
+calls this same command as part of a bigger sync flow) — use this skill directly only when you
+need the focus map for your own immediate work, not to persist knowledge.
+
+## Running it
+
+```bash
+obsidian-wiki code-understand --project <dir> [--backend auto|builtin|codegraph] \
+    [--changed <file>...] [--since <sha>] [--max-symbols N] [--pretty]
+```
+
+- `--project` — defaults to the current directory.
+- No `--changed`/`--since` — seeds from every tracked file (full-project scan).
+- `--changed <file>` (repeatable) — seed from specific files you already know are relevant
+  (e.g. files in the diff you're about to make).
+- `--since <sha>` — seed from everything changed since a git ref (e.g. `--since HEAD~5` or a
+  base branch).
+- `--backend` — `auto` (default) uses CodeGraph when installed, falls back to the builtin
+  extractor otherwise; force `builtin` for the dependency-free path, or `codegraph` to require
+  the enhanced backend (fails loudly if it's not installed rather than silently degrading).
+- `--max-symbols` — cap the focus map size (default 50); keep this small for a quick
+  orientation pass, raise it for a thorough one.
+
+**GUARD:** if the command is unavailable or errors, skip it and fall back to your normal
+exploration (grep/read) — this is an accelerant, not a dependency. Don't block work on it.
+
+## Reading the output
+
+1. **Evidence type matters.** When `backend: codegraph`, focus-map entries are structural
+   facts (real call-graph edges) — cite them directly. When `backend: builtin`, `defines`/
+   `imports`/`changed-file` entries are facts, but `rg-reference` entries are text-match
+   evidence only — open the file and confirm before treating it as a real relationship.
+2. **Open what it cites, nothing else first.** The focus map tells you *where* to read; it
+   never contains source bodies. Go read those `file:line` locations before forming an opinion
+   on the architecture.
+3. **Cite it in your own output.** If you explain the architecture back to the user or write
+   it into a PR description/commit message, keep the `(file:lines)` citations from the focus
+   map or from the source you opened — don't assert structure without a pointer to it.
+4. **It's a cache, not a deliverable.** Never write the raw JSON or `.codegraph/` into the
+   wiki vault, a PR description, or committed docs — it's a disposable sidecar in the project
+   repo (git-ignored). Re-run it fresh next time rather than treating old output as durable.
+
+## If CodeGraph isn't installed
+
+`auto`/`builtin` still work with zero setup (uses the regex AST extractor + `rg`). If you want
+the enhanced cross-file call-graph evidence and the user is fine with installing something:
+
+```bash
+npm install -g @colbymchenry/codegraph
+```
+
+or point `CODE_UNDERSTANDING_CODEGRAPH_BIN` at an existing binary. Never install without the
+user's go-ahead — ask first, then re-run the command.
+
+Check `obsidian-wiki doctor --project <dir>` for the `code-understanding.*` capability lines
+(`builtin`, `rg`, `codegraph`, `codegraph-index`, `codegraph-fresh`) to see what's available
+before deciding which backend to request explicitly.
--- a/.skills/llm-wiki/SKILL.md
+++ b/.skills/llm-wiki/SKILL.md
@@ -590,6 +590,8 @@ The wiki is configured through environment variables (see `.env.example`). The o
 - `OBSIDIAN_LINK_FORMAT` — Internal link syntax: `wikilink` (default) or `markdown`
 - `WIKI_TOKEN_WARN_THRESHOLD` — Emit a warning in `wiki-status` when the full-wiki token estimate exceeds this value (default: `100000`). Set to `0` to disable. See `wiki-status` for the token footprint report.
 - `WIKI_STAGED_WRITES` — When `true`, all LLM-written pages go to `_staging/<category>/` for human review before promotion. See `wiki-setup` and `wiki-stage-commit` for details.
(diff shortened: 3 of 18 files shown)


Ar9av:
looks good
Also pushed a small follow-up commit adding a standalone `.skills/code-understand/SKILL.md` so the focus-map extractor can be invoked directly by any coding agent on demand, not just as the hidden `wiki-update`

mike840609:
Here's the`/wiki-update` ingest flow for the new project. The other data retrieval steps remain unchanged.


image

Issue #152, opened 2026-07-25
daily-update Setup Mode references scripts/ (daily-update.sh, plist, wiki-notify.sh) missing from the published wheel


BigLep:
## Summary

The `daily-update` skill's **Setup Mode** references helper scripts that are present in the git repo under [`scripts/`](https://github.com/Ar9av/obsidian-wiki/tree/main/scripts) but are **not included in the published PyPI wheel**. As a result, Setup Mode cannot run as written.

This is the same class of packaging gap as #143 (the `wiki-stop-capture.sh` hook), which is now closed — but that fix did not cover the `scripts/` directory.

## Details

`daily-update/SKILL.md` (Setup Mode) references:

- `$OBSIDIAN_WIKI_REPO/scripts/daily-update.sh`
- `$OBSIDIAN_WIKI_REPO/scripts/com.obsidian-wiki.daily-update.plist`
- `$OBSIDIAN_WIKI_REPO/scripts/wiki-notify.sh`

In the installed wheel, `obsidian_wiki/_data/` contains `skills/`, `bootstrap/`, and `.env.example`, but **no `scripts/` directory**:

```
$ find ~/.local/share/uv/tools/obsidian-wiki -name 'daily-update.sh' -o -name 'wiki-notify.sh' -o -name 'com.obsidian-wiki.daily-update.plist'
# (no output)

$ ls .../site-packages/obsidian_wiki/_data/skills/daily-update/
SKILL.md          # <- only the SKILL.md, no scripts/
```

Meanwhile the repo ships them in git:

```
$ gh api repos/Ar9av/obsidian-wiki/contents/scripts --jq '.[].name'
com.obsidian-wiki.daily-update.plist
daily-update.sh
extract-jsonl.py
manifest.py
wiki-notify.sh
```

## Environment

- obsidian-wiki **2026.7.9** (installed via `uv tool`, Python 3.12)
- macOS

## Impact

Running `daily-update` Setup Mode ("set up the daily cron" / "install terminal notification") fails at Step 1, which checks that `scripts/daily-update.sh` exists. Users cannot install the launchd cron or the terminal-notification (`wiki-notify.sh`) infrastructure from a pip/uv install without manually fetching the scripts from GitHub.

## Suggested fix

[...]

Issue #143, opened 2026-07-17
wiki-setup Stop hook references .claude/hooks/wiki-stop-capture.sh, which is missing from the published wheel


BigLep:
## Summary

The `wiki-setup` skill's "Install the Stop Hook" step wires `~/.claude/settings.json` to run `<REPO_PATH>/.claude/hooks/wiki-stop-capture.sh`, but that script is **not included in the published pip/uv wheel**. It exists in the git repo but is stripped from the distribution, so anyone who installed via `uv tool install` / `pip install` has a Stop hook pointing at a file that doesn't exist.

## Environment

| Component | Version |
|---|---|
| obsidian-wiki | 2026.7.7 |
| Install method | `uv tool install` (`~/.local/share/uv/tools/obsidian-wiki`) |
| Claude Code | 2.1.205 |
| Python | 3.9.2 |
| OS | macOS 26.3.1 (Darwin 25.3.0, arm64) |

## Steps to reproduce

1. Install obsidian-wiki as a tool: `uv tool install obsidian-wiki` (version 2026.7.7).
2. Run the CLI setup so skills + config are installed.
3. Run the `/wiki-setup` skill to initialize a vault, and accept the optional **"Want to auto-capture findings at session end?"** Stop-hook step.
4. The skill instructs the agent to merge this into `~/.claude/settings.json` (SKILL.md around line 209):
   ```json
   { "type": "command", "command": "bash <REPO_PATH>/.claude/hooks/wiki-stop-capture.sh" }
   ```
   where `<REPO_PATH>` resolves to the installed package's `_data` dir (`OBSIDIAN_WIKI_REPO`).

## What happened

The referenced script is absent from the installed package. The distribution only bundles:

```
obsidian_wiki/_data/skills/
obsidian_wiki/_data/bootstrap/
obsidian_wiki/_data/.env.example
```

There is no `.claude/hooks/` directory in `_data`, and a system-wide search finds nothing:

```console
$ find ~/.local/share/uv/tools/obsidian-wiki -name 'wiki-stop-capture*'
# (no output)

$ grep -c '\.claude/hooks' \
    ~/.local/share/uv/tools/obsidian-wiki/lib/python3.9/site-packages/obsidian_wiki-2026.7.7.dist-info/RECORD
0
```

[...]

PR #180, opened 2026-08-17
docs(vault-skill-factory): resolve skill-creator scripts path for both install layouts

Lupin133:
## What

`vault-skill-factory` Step 4 points at `$OBSIDIAN_WIKI_REPO/.skills/skill-creator/scripts/`. That
path only exists for a source checkout.

`cli.py` `skills_dir()` resolves to `<pkg>/_data/skills` for a wheel install and
`<repo>/.skills` for a checkout, and `write_config()` sets `OBSIDIAN_WIKI_REPO = skills_dir().parent`.
So the skills root is `$OBSIDIAN_WIKI_REPO/skills` under a wheel (no leading dot — `pyproject.toml`
force-includes `".skills"` to `"obsidian_wiki/_data/skills"`) and `$OBSIDIAN_WIKI_REPO/.skills` from
source. No single literal works for both.

Verified by building the wheel and installing it into a clean venv:

```
$ python -c "from obsidian_wiki.cli import skills_dir; sd=skills_dir(); \
    print((sd.parent/'.skills'/'skill-creator'/'scripts').is_dir(), \
          (sd/'skill-creator'/'scripts').is_dir())"
False True
```

Sibling references like `$OBSIDIAN_WIKI_REPO/scripts/...` (claude-history-ingest, daily-update) are
unaffected — `"scripts"` force-includes to `obsidian_wiki/_data/scripts`, so that literal is correct
in both layouts. This is the same class as #143 and #152, in a file those fixes did not touch.

## How

List both layouts and take the first that exists, mirroring the phrasing `wiki-setup` already uses
for the `wiki-stop-capture.sh` lookup, rather than swapping one broken literal for another.

Docs only, no behaviour change.

### DIFF
--- a/.skills/vault-skill-factory/SKILL.md
+++ b/.skills/vault-skill-factory/SKILL.md
@@ -109,6 +109,8 @@ checkout, so check both layouts under `$OBSIDIAN_WIKI_REPO` and use the first th
   at the bundled `_data/` dir, which ships the skills under `skills/`).
 - `$OBSIDIAN_WIKI_REPO/.skills/skill-creator/scripts/` — source checkout.
 
+Whichever resolves, it holds:
+
 - `improve_description.py` — tighten the generated `description` for better triggering.
 - `package_skill.py` — bundle the skill dir into a distributable archive.
 - `quick_validate.py` — sanity-check the skill's structure.
--- a/.skills/vault-skill-factory/SKILL.md
+++ b/.skills/vault-skill-factory/SKILL.md
@@ -102,7 +102,13 @@ one-line-install convention) — still **not** an install, just a manifest.
 
 ## Step 4: Optionally lean on skill-creator
 
-`skill-creator` ships reusable scripts (`$OBSIDIAN_WIKI_REPO/.skills/skill-creator/scripts/`):
+`skill-creator` ships reusable scripts. Their path differs between a pip/uv install and a source
+checkout, so check both layouts under `$OBSIDIAN_WIKI_REPO` and use the first that exists:
+
+- `$OBSIDIAN_WIKI_REPO/skills/skill-creator/scripts/` — packaged install (`OBSIDIAN_WIKI_REPO` points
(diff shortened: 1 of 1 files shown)


Ar9av:
Verified the claim rather than taking it on trust. Built the wheel and listed it:

```
$ python3 -m build --wheel
$ python3 -c "import zipfile,glob; z=zipfile.ZipFile(glob.glob('*.whl')[0]); \
    print([x for x in z.namelist() if 'skill-creator/scripts' in x][:2])"
['obsidian_wiki/_data/skills/skill-creator/scripts/__init__.py',
 'obsidian_wiki/_data/skills/skill-creator/scripts/aggregate_benchmark.py']
```

`skills_dir()` (cli.py:67) returns `<pkg>/_data/skills` for that wheel, `write_config()` (cli.py:334) sets `OBSIDIAN_WIKI_REPO = skills_dir().parent`, so the packaged path is `$OBSIDIAN_WIKI_REPO/skills/...` and the checkout path is `$OBSIDIAN_WIKI_REPO/.skills/...`. The old literal only ever resolved from source. Diagnosis holds.

Checking both layouts is also the right call over picking one literal, and reusing the `wiki-setup` phrasing (SKILL.md:225) keeps the two lookups readable side by side.

One thing I fixed and pushed (706e455): the two layout bullets and the three script-name bullets sat in one list separated by a blank line, which markdown renders as a single five-item list. A reader scanning it can't tell locations from filenames. Added a `Whichever resolves, it holds:` lead-in between them.

Scoped it while I was in there: the sibling `$OBSIDIAN_WIKI_REPO/scripts/...` references in `claude-history-ingest` and `daily-update` are fine, since `"scripts"` force-includes to `_data/scripts`. `wiki-context-pack` (SKILL.md:65) references `$OBSIDIAN_WIKI_REPO/obsidian_wiki/cli.py`, which a wheel install won't have, but that one sits behind an `elif` guard with a working fallback ahead of it, so nothing breaks. Separate issue if you want it tidied.

Good catch on a file #143 and #152 missed.

Issue #113, opened 2026-06-21
Question: Why delete after ingest?


ethanopp:
Curious as to why the ingest function deletes from _raw once processed? Karpathy’s initial write up mentions raw as “immutable”, and don’t we lose any way to recover, reingest, or even just reference the original cited source in the future if we delete?

Ar9av:
Good question. It points at a naming collision between two different things called "raw."

Karpathy's "Layer 1: Raw Sources (immutable)" means your original documents: PDFs, articles, conversation logs sitting outside the vault in `OBSIDIAN_SOURCES_DIR`. The system never touches those.

The `_raw/` folder inside the vault is something else: a staging inbox for quick captures and drafts waiting to be distilled into a proper wiki page. Once `wiki-ingest` promotes a file from `_raw/` into `concepts/`, `skills/`, etc., it deletes the original. The skill does this so the next ingest run won't reprocess the same file (`.skills/wiki-ingest/SKILL.md:70`).

Provenance mostly survives. The promoted page inherits the `sources:` field from the `_raw/` file's own frontmatter, so the citation trail back to the original external source stays intact. The draft text itself doesn't survive, though. For content where `_raw/` held the only copy, like a quick-capture finding typed straight from a conversation with no external document behind it, that's a real loss if the promoted page drops detail.

The vault setup already creates an `_archives/` folder, but nothing currently uses it for promoted `_raw/` files. Moving them there instead of deleting them would close the gap without breaking the duplicate-processing guard. Fixing this now in `wiki-ingest` — will follow up here once it's in.

Ar9av:
Fixed. `wiki-ingest` now moves a promoted `_raw/` file into `_raw/_archived/` instead of deleting it. The duplicate-processing guard still holds since the file leaves the top-level `_raw/` scan, but the original draft survives for cases where it's the only copy. Also updated `wiki-status` (pending-drafts count excludes `_archived/`), `wiki-setup`, and `llm-wiki` to call out the distinction between this staging folder and Karpathy's immutable Layer 1 sources, so the two don't get conflated again.

Tested the move mechanics directly in a sandbox before merging.

Ar9av:
Merged the fixes in : https://github.com/Ar9av/obsidian-wiki/commit/ec2b6bc101b9d97fe27a5aff496d79e7df4b1649

LGDHuaOPER:
How should I write the path for `OBSIDIAN_SOURCES_DIR`?
[...]

PR #181, opened 2026-08-17
docs(env): promoted _raw/ files are archived, not deleted

Lupin133:
## What

`.env.example`'s `OBSIDIAN_RAW_DIR` comment still says:

```
# Promoted files are deleted from _raw/ after processing.
```

That is the pre-#113 behaviour. Three skills now document the opposite:

- `.skills/wiki-ingest/SKILL.md` — "After promoting a file to a proper wiki page, **move the original
  into `_raw/_archived/`** ... instead of deleting it."
- `.skills/wiki-status/SKILL.md` — "excluding the `_archived/` subdirectory (already-promoted drafts
  kept for provenance, not pending work)".
- `.skills/llm-wiki/SKILL.md` — "`wiki-ingest` still moves rather than deletes them on promotion."

`.env.example` is usually the first file someone opens to configure the tool, so it is the worst
place to still promise that a quick-capture draft gets destroyed — which is the outcome #113 was
filed about.

## How

One comment line. Docs only, no behaviour change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -205,5 +205,5 @@ CODE_UNDERSTANDING_CODEGRAPH_BIN=
 # A directory inside OBSIDIAN_VAULT_PATH for unprocessed draft pages.
 # Drop rough notes, clipboard pastes, or quick captures here.
 # wiki-ingest will promote them to proper wiki pages when you run it.
-# Promoted files are deleted from _raw/ after processing.
+# Promoted files are moved into _raw/_archived/ after processing, not deleted.
 OBSIDIAN_RAW_DIR=_raw


Ar9av:
Verified the claim against the tree. `.env.example:208` was the last place still promising deletion; the three skills the PR cites (`wiki-ingest`, `wiki-status`, `llm-wiki`) all describe the archive behavior, and `ec2b6bc` is where it changed. `obsidian_wiki/lint.py`, `trust.py`, `context_pack.py`, `graph_analysis.py` and `batch.py` all skip `_archived` in their scans, so the code agrees too. No other doc contradicts it: `docs/configuration.md` mentions `_raw/` promotion without claiming files get removed.

One thing left on the table, not a blocker: `docs/configuration.md:184` says the directory is created by `wiki-setup` and configurable via `OBSIDIAN_RAW_DIR`, but never tells you where a promoted draft ends up. Someone reading only the docs page still won't know `_archived/` exists. Worth a follow-up sentence there.

LGTM.

Issue #188, opened 2026-08-23
fix(cli): tolerate GBK consoles during setup output


georgelichen:
## Summary

`obsidian-wiki setup` crashes before skill installation when Windows Python uses a GBK console encoding.

## Reproduction

In a PowerShell session whose Python stdout encoding is `gbk`, run:

    obsidian-wiki setup --vault C:\\path\\to\\vault

The command writes the config, then raises:

    UnicodeEncodeError: 'gbk' codec can't encode character '\\u2705'

The CLI prints check marks, box drawing characters, warning symbols, arrows, and other Unicode status text. A narrow Windows console encoding should not make setup abort.

## Expected behavior

Setup should continue and use a replacement-safe console error policy when the active output encoding cannot represent a status character. Structured output and file contents should remain unchanged.

## Proposed validation

Add a regression test with a strict GBK text stream, invoke the CLI output setup, and verify printing a check mark does not raise `UnicodeEncodeError`.

Ar9av:
Fixed by #189 (`_configure_console_output()` sets `errors="replace"` on stdout/stderr).

Verified on a real GBK Windows box rather than a simulated stream — Windows Server 2022 with the system locale set to `zh-CN` (ACP 936), so a redirected `stdout` genuinely reports `encoding=gbk`:

- `obsidian-wiki setup --vault C:\Users\...\wiki` → exit 0, setup completes, `doctor` passes; check marks / box drawing / arrows degrade to `?` instead of raising.
- Confirmed both as an administrator (symlink install) and as a non-admin user (WinError 1314 → copy fallback).
- The released PyPI build `2026.8.7` already contains the fix.

Closing.

Issue #187, opened 2026-08-23
fix(cli): fall back to copied skills without Windows symlink privilege


georgelichen:
## Summary

`obsidian-wiki setup` still aborts on Windows when the current user cannot create symbolic links.

## Reproduction

On a standard Windows account without Developer Mode or the Create symbolic links user right:

1. Run `obsidian-wiki setup`.
2. Config and vault verification succeed.
3. Global skill installation fails at `Path.symlink_to(...)` with:

    OSError: [WinError 1314] A required privilege is not held by the client.

The existing `--copy` option works around the problem, but the default setup mode does not explain the fallback or automatically use it.

## Expected behavior

When symlink creation fails specifically with Windows error 1314, setup should warn that symbolic links are unavailable and continue by copying the bundled skill directory. It should preserve the current behavior for unrelated filesystem errors and keep `--copy` as an explicit deterministic mode.

## Proposed validation

Add a regression test that simulates WinError 1314 from `Path.symlink_to`, verifies the skill is copied, and verifies that unrelated `OSError` values are still raised.

Ar9av:
Fixed on `main` in 7620af7 (PR #189) — `install_skills()` now falls back to copying skills when `symlink_to` raises WinError 1314, re-raising any other `OSError` unchanged, with regression coverage in `tests/test_cli_install.py`. Closing.

PR #189, opened 2026-08-23
fix(cli): stabilize Windows skill setup

### LINKED ISSUE #186: fix(cli): handle Windows directory junctions during skill installation
## Summary

obsidian-wiki setup crashes on Windows when an existing global skill entry is a directory junction (NTFS reparse point).

## Reproduction

Environment:

- Windows 10/11
- Python 3.12+
- obsidian-wiki 2026.8.4 (or current main)

1. Have a global skill mirror such as C:\\Users\\<user>\\.claude\\skills\\agent-reach that is a directory junction to another skill directory.
2. Run obsidian-wiki setup.
3. Setup writes the config and verifies the vault, then fails while reinstalling global skills:

    OSError: Cannot call rmtree on a symbolic link

## Root cause

obsidian_wiki/cli.py lines 129-135 check Path.is_symlink() and then treat the path as a normal directory when that check is false. On Windows, Path.is_symlink() is false for a directory junction while Path.is_junction() is true. Because the junction target contains SKILL.md, the code calls shutil.rmtree(link_path), which intentionally refuses to recurse through a link or reparse point.

The observed state is reproducible with:

    Path(path).is_symlink()  # False
    Path(path).is_junction() # True
    Path(path).is_dir()      # True

## Expected behavior

Treat directory junctions and reparse points like managed links: remove the link itself without traversing its target, then recreate the skill link. Real copied skill directories should retain the existing replacement behavior, and unmanaged directories should still be skipped.

## Related work

- PR #19 fixes Git-on-Windows pseudo-symlink text files in setup.sh; it does not cover the Python CLI or directory junctions.
- PR #164 concerns Windows setup stability but does not modify install_skills.

## Proposed validation

Add a Windows regression test that creates a temporary directory junction, runs install_skills, and verifies that the junction is replaced without deleting the target directory. Keep the test skipped on platforms without junction support.

georgelichen:
## Summary

- Fix global skill installation when an existing Windows entry is a directory junction.
- Fall back to copied skills only when Windows rejects symlink creation with `WinError 1314`.
- Make CLI status output replacement-safe for GBK PowerShell consoles.
- Add regression coverage for junction replacement, privilege fallback, unrelated symlink errors, and console encoding.

Fixes #186
Related: #187, #188

## Root Cause

Windows `Path.is_symlink()` does not identify directory junctions, so the installer previously sent a junction to `shutil.rmtree()`. Standard Windows accounts can also lack the privilege required by `Path.symlink_to()`. Finally, strict GBK stdout encoding caused status icons to abort setup before installation.

## Test Plan

- [x] `python -m pytest -q tests/test_cli_install.py tests/test_ast_extractor.py tests/test_cache.py tests/test_graph_analysis.py tests/test_graphrag.py` (196 passed)
- [x] `python -m py_compile obsidian_wiki/cli.py`
- [x] `git diff --check`
- [x] Ran `obsidian-wiki setup --vault C:\Users\lichen\source\repos\TFX-DMS-Wiki\wiki` on Windows; exit code 0 and 40 skills installed per global agent directory
- [ ] Full suite: existing Windows doctor/environment failures remain outside this change

### DIFF
--- a/obsidian_wiki/cli.py
+++ b/obsidian_wiki/cli.py
@@ -15,6 +15,7 @@ import json
 import os
 import re
 import shutil
+import stat
 import sys
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Ar9av:
Thanks @georgelichen for fixing windows issues

PR #190, opened 2026-08-23
fix(graphrag): strip trailing punctuation on the gap and list query paths

stemForge:
## Problem

`classify_query()` strips punctuation on the `direct` path but not on the `gap`
and `list` paths. Both of those end in `.strip().split()`, and `.strip()` without
arguments removes only whitespace — so a trailing `?` survives on the extracted
term.

The result is that a gap query returns nothing even when the page exists:

```python
from obsidian_wiki.graphrag import classify_query

classify_query("What do I know about mise?")   # ('gap', ['mise?'])   <- matches nothing
classify_query("What do I know about mise")    # ('gap', ['mise'])    <- finds the page
classify_query("What is mise?")                # ('direct', ['mise']) <- direct path is fine
```

`mise?` matches no title, tag or summary, so `query()` comes back with an empty
candidate list. Since gap questions nearly always end in a question mark, this
path was broken for its ordinary input.

## Fix

Both paths now go through a shared `_split_terms()` helper that splits on
whitespace and strips the same punctuation set the `direct` path already used,
now named `_TERM_PUNCT`.

The `direct` path keeps its behaviour unchanged — its `len(w) > 2` guard still
applies to the unstripped token, so only the inline literal is replaced by the
constant.

## Tests

Three regression tests in `TestClassifyQuery` cover the gap path, the list path,
and agreement between the gap and direct paths on the same subject. All three
fail without the fix.

Full suite: 651 passed, 1 skipped (648 before, +3 new).

## Note

While investigating this I ran into a second, larger issue in the same function:
`_score()` matches substrings without word boundaries, and the stop-word set is
English-only, so non-English questions about absent topics return a false
positive flagged `index_only: true` rather than an empty list. That one changes
ranking behaviour, so I left it out of this PR — happy to open a separate issue
with the measurements if that's useful.

### DIFF
--- a/obsidian_wiki/graphrag.py
+++ b/obsidian_wiki/graphrag.py
@@ -326,6 +326,16 @@ _STRUCTURAL = (
 )
 
 
+# Trailing punctuation to strip off extracted terms. A question mark left on a
+# token means it can never match a title, tag or summary.
+_TERM_PUNCT = "?,.'\""
+
+
+def _split_terms(text: str) -> list[str]:
+    """Split on whitespace, strip surrounding punctuation, drop empties."""
+    return [t for t in (w.strip(_TERM_PUNCT) for w in text.split()) if t]
+
+
 def classify_query(question: str) -> tuple[str, list[str]]:
     """Return (answer_type, extracted_terms).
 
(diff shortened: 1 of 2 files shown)


stemForge:
Hello @Ar9av,

thanks for building this. I've been setting up a vault with it over the last few weeks, and it has genuinely changed how I keep notes. The three-layer pattern and the way skills route into it are a pleasure to work with.

I opened this and two issues (#191 , #192 ) along the way. 

These are my first contributions here, so if the form isn't what you expect (scope, commit style, level of detail, or whether something should have been a discussion instead of an issue) please just say so. I'd welcome the feedback and will adjust.

Ar9av:
@stemForge this lgtm, thanks for the PR and contributing
I dont have windows OS to test thoroughly, could you help me verfiy the issues you have posted as I post and try to replicate it on a sandbox windows

stemForge:
@Ar9av Thanks for merging, and for the quick fix in #193.

I think there's a misunderstanding though: I'm on a Mac and my issues #191 and #192 aren't platform-specific. I do have a Windows PC, but with no dev environment or AI tooling on it, so I can't test there.

Either way, I'd like to keep contributing where I can, just let me know.

Issue #177, opened 2026-08-17
wiki-lint Check 2 documents a naive wikilink rule; --consolidate Action 1 then rewrites valid links


Lupin133:
## Description

`wiki-lint`'s Check 2 documents a link-resolution rule that is more naive than the one
`obsidian_wiki/lint.py` already implements, and `--consolidate` then acts destructively on its output.

`.skills/wiki-lint/SKILL.md` lines 50-57:

```
### 2. Broken Wikilinks
Find `[[wikilinks]]` that point to pages that don't exist.
**How to check:**
- Grep for `\[\[.*?\]\]` across all pages
- Extract the link targets
- Check if a corresponding `.md` file exists
```

`.*?` captures the whole inside of the brackets, so the extracted target keeps everything Obsidian
treats as non-filename: `[[page|Display]]`, `[[page#Heading]]`, `[[page#^blockid]]`, and `[[page\|A]]`
(the escape a Markdown table cell requires). `lint.py` line 43 already gets the first three right —
`\[\[([^\]|#]+?)(?:[|#][^\]]*?)?\]\]` excludes `|` and `#` from the capture. An agent following the
SKILL.md prose does not.

This matters because Action 1 in `--consolidate` mode (lines 483-489) consumes Check 2's list and
writes to the vault:

```
- If a unique best match exists (edit distance <= 2 characters or same root word): rewrite the link.
- If no match or ambiguous: convert to plain text (`~~[[Target]]~~` -> `Target`) and add a comment
```

`[[page#Notes]]` is within edit distance 2 of `page`, so it is silently rewritten and the heading
anchor is lost. Longer anchors and aliases fall to the second branch and become plain text. The
dry-run preview shows these as legitimate repairs, so approving it is the natural thing to do.

Check 12 (line 317) has the same gap on `relationships[].target`: it strips `[[`/`]]` and `.md` and
normalises case and spaces, but never strips `|alias`, `#heading` or `^block`.

## Steps to Reproduce

`python repro.py` — implements Check 2 exactly as written.

```python
import os, re, shutil, tempfile

VAULT = tempfile.mkdtemp(prefix="wikilint-repro-")
PAGES = {
    "notes/alpha.md": "# Alpha\n\n## Section\n\nSome text. ^blockid\n",
    "notes/beta.md": "# Beta\n",
    "dashboard.base": "filters: {}\n",
    "notes/index.md": """# Index

Alias link:        [[alpha|Alpha display]]
[...]

Issue #176, opened 2026-08-17
lint_vault reports image/PDF/canvas/.base embeds and table-escaped pipes as broken links


Lupin133:
## Description

`lint_vault()` reports every wikilink whose target is not a `.md` page as a broken link. Obsidian
vaults routinely link to attachments and non-markdown files, so on a real vault `broken_links` fills
with entries that are not broken.

Two causes, both in `obsidian_wiki/lint.py` at 5cab5c1:

1. `_iter_pages()` (lines 56-60) globs `vault.rglob("*.md")`, so `by_slug` only ever contains
   markdown stems. `_parse_page()` (lines 162-165) collects **every** `_WIKILINK_RE` hit, including
   embeds. `![[diagram.png]]`, `![[paper.pdf]]`, `[[board.canvas]]` and `[[dashboard.base]]` are
   therefore compared against an index that structurally cannot contain them.

2. `_WIKILINK_RE` (line 43) is `\[\[([^\]|#]+?)(?:[|#][^\]]*?)?\]\]`. Excluding `|` from the capture
   handles `[[page|alias]]` correctly, but inside a Markdown table the pipe must be written `\|`, and
   the backslash **is** in the allowed set — so `[[beta\|B]]` captures `beta\` (trailing backslash)
   and never resolves.

Separately, `[[alpha.md]]` fails too: `_normalise_node_id()` strips a `.md` suffix (line 149-150) but
the link list built in `_parse_page()` uses `_slug(raw.split("/")[-1])` with no suffix stripping, so
the explicit-extension form does not match the page's slug.

## Steps to Reproduce

Save as `repro.py` at the repo root and run `python repro.py .`

```python
import shutil, sys, tempfile
from pathlib import Path
sys.path.insert(0, sys.argv[1] if len(sys.argv) > 1 else ".")
from obsidian_wiki.lint import lint_vault

VAULT = Path(tempfile.mkdtemp(prefix="ow-lint-repro-"))
FM = """---
title: {title}
type: concept
summary: repro fixture
tags: [repro]
sources: []
created: 2026-08-17
updated: 2026-08-17
base_confidence: 0.5
lifecycle: draft
lifecycle_changed: 2026-08-17
provenance: extracted
relationships: []
---
"""
(VAULT / "notes").mkdir(parents=True)
(VAULT / "attachments").mkdir(parents=True)
[...]

PR #198, opened 2026-08-27
feat(lint): report pages that share a slugged stem

zorionarrillaga:
A page can disappear from search and nothing reports it.

Two pages whose filenames slug to the same stem are one node to `graph_analysis`, and one entry to
`graphrag`. Measured on a three-page vault holding `concepts/vector-search.md` ("The retrieval
technique") and `entities/vector-search.md` ("Pinecone (the product)"):

```
graphrag.build_index(vault)  ->  slugs ['embeddings', 'vector-search']
   'vector-search'           ->  'Pinecone (the product)'  entities/vector-search.md
```

`concepts/vector-search.md` is absent from the index that `memory_search` and `graph-query` read. The same merge runs through `god_nodes`, `detect_communities`,
`betweenness_centrality` / `bridge_pages`, `shortest_path`, `neighborhood` and `_insights.md`.

`duplicate_titles` does not cover it, because it keys on the `title:` value while identity keys on
the stem — the two pages above have different titles.

## Where the collision comes from

`graph_analysis._page_slug(path, root)` returns `_slug(path.stem)` (`graph_analysis.py:64-65`), so
identity is the slugged filename. `write_page` builds `f"{_slug(category)}/{_slug(title)}.md"`
(`server.py:77`); `category` is a required parameter of `write_page` (`server.py:69`), exposed as the
`memory_write` MCP tool (`server.py:160`) and over HTTP, where `PageWrite` defaults it to `concepts`
(`server.py:141`).

Two `write_page` calls into a fresh vault, nothing else:

```python
write_page("Vector Search", "concepts", "The technique.")
write_page("vector-search", "entities", "The Pinecone product.")
```

```
files written      concepts/vector-search.md   entities/vector-search.md   log.md
graph              2 pages -> 1 node  ['vector-search']       (root log.md is graph-excluded)
duplicate_titles   []
duplicate_stems    [{'stem': 'vector-search',
                     'pages': ['concepts/vector-search.md', 'entities/vector-search.md']}]
```

A hand-authored vault reaches it too: `Vector Search.md` beside `vector-search.md` slugs to one stem,
in one folder or two.

## The change

Build a stem index over `graph_analysis.iter_pages` using `graph_analysis._page_slug`, and report
stems holding more than one page. Both are borrowed from that module rather than reimplemented, so
the report keys as the merge does. It reads pages rather than links, so a collision nothing
references is still reported.

Severity is `warn`, beside `duplicate_titles`. **Migration:** a vault carrying a collision moves
`pass` → `warn`, so `obsidian-wiki lint` stays exit 0 while `obsidian-wiki lint --strict` becomes
exit 1. If you run `--strict` in CI, an existing collision will surface there on upgrade.

Six tests in `tests/test_lint.py`. Suite: **687 passed, 5 skipped (681 before, +6 new)**. Each test
fails against unpatched `lint.py`. Mutants run at the call sites this patch adds, all caught:
dropping either sort; `> 1` → `> 0`; `iter_graph_pages` → a raw `rglob`; the borrowed slug → a raw
`Path.stem`; the borrowed slug → `stem.lower()`; restricting to different folders; and removing the
key from the `warn` tuple.

## Two things worth your call rather than mine

**`_bootstrap/` becomes visible to this one check.** `lint.SKIP_DIRS` (`lint.py:19`) skips
`_bootstrap` and `.git`; `graph_analysis.SKIP_DIRS` skips `_meta` and `_readouts` instead. Borrowing
the graph's selection is what makes the report match the merge, but it does mean a collision *inside*
`_bootstrap/` is now reported, and `test_lint_excludes_bootstrap_scaffolding_from_page_health`
(`tests/test_trust.py:840`) exists because you wanted that directory out of page health. That test
still passes — its one file has a unique stem — so this is a question rather than a breakage: keep
the graph's selection, or intersect it with `lint.SKIP_DIRS` and lose collisions that the graph does
merge. One line either way; I did not want to pick it quietly.

**`lint`'s own `by_slug` is stem-keyed too** (`lint.py:214`), so a collision the graph excludes — two
pages inside `_meta/`, say — can still absorb one page's in-links into the other and affect
`orphan_pages`. This patch does not touch that, and does not report it.

## What I did not change

- **Page identity.** Making `_page_slug` path-aware would change what `[[name]]` resolves to and what
  a stored `_insights.md` snapshot means. This reports the condition instead of deciding it.
- **`.skills/wiki-lint/SKILL.md`**, which enumerates the checks. #177 is open on that file's
  divergence from `lint.py`, so I left it alone — with the consequence that an agent running the
  skill will not surface this finding until the spec catches up.
- **Link collection.** #176 and #177 concern `_parse_page`'s wikilink harvest; this check reads no
  links.

`wiki-dedup` is the neighboring tool and does not overlap: it scores candidates on title similarity,
and it exists to *merge* pages that are the same concept. The case here is two pages that are
legitimately different and should stay separate while the graph merges them anyway. Its own
`node_id` is the vault-relative path, which is the identity this check is pointing at.

## Limits

`write_page` was executed with `fastapi`, `pydantic` and `mcp` stubbed — it imports none of them, but
the server was not run for real. The vaults above were constructed; I have not seen this in a vault
of yours, and it does not need observing: two `memory_write` calls with different categories produce
it mechanically. On Python 3.12 the page-ordering half of one test may be satisfied by the walk order
rather than by the sort; the assertion holds either way, and I ran 3.9, 3.11 and 3.14.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -54,6 +54,26 @@ obsidian-wiki lint --allow-lifecycle active --allow-relationship-type synthesize
 
 Lint resolves its vault and schema together: explicit path (no config inheritance), positional `@name`, nearest CWD `.env`, then global config. CLI schema flags extend/replace that resolved vault's settings and are recorded in the JSON `schema` block.
 
+### Colliding page stems
+
+`duplicate_stems` reports two or more pages whose filename stems are equal after slugging —
+`concepts/vector-search.md` and `entities/vector-search.md` across folders, or `Vector Search.md`
+beside `vector-search.md` inside one. The graph
+identifies a page by its bare stem, so those pages are a single node in every graph metric:
+degree, communities, betweenness, and the `graph-analyse --around` blast radius. The check keys
+on the same slug and the same page selection `graph_analysis` uses, so what it reports is exactly
+what the graph will merge — root `index.md`/`log.md`/`hot.md`/`_insights.md` are excluded, and a
+legitimate `concepts/index.md` is not.
+
+It is independent of how any link is written: the collision is between the pages, not between
+references to them, so a vault where nothing links to the stem is still reported.
+
+`duplicate_titles` does not cover this — it keys on the `title:` value, and stems can collide
+while titles differ (`"Vector Search"` and `"vector-search"` slug to the same stem).
+
+The check warns. A vault carrying a collision moves from `pass` to `warn`, which leaves
+`obsidian-wiki lint` at exit 0 and takes `obsidian-wiki lint --strict` to exit 1.
+
 ### Lifecycle transition checking
 
 `illegal_lifecycle_transitions` compares each page's current `lifecycle` against the value recorded in `_meta/trust-ledger.json` at its last review, and flags moves the state machine forbids: any state falling back to `draft` (only ingest sets `draft`), and any exit from `archived` (a restore is a deliberate delete-and-recreate, not a transition).
--- a/obsidian_wiki/lint.py
+++ b/obsidian_wiki/lint.py
@@ -8,6 +8,8 @@ from collections.abc import Collection
 from pathlib import Path
 from typing import Any
 
+from obsidian_wiki.graph_analysis import _page_slug as graph_page_slug
+from obsidian_wiki.graph_analysis import iter_pages as iter_graph_pages
 from obsidian_wiki.trust import (
     ALLOWED_LIFECYCLES,
     TRUST_LEDGER_RELATIVE_PATH,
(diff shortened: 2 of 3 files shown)


Ar9av:
Tested on st3ve (fresh clone, PR head `f085d14`).

- Full suite: 688 passed, 4 skipped. My venv was missing the `tree-sitter`/`igraph` extras, which accounts for the skip-count gap from the PR's 5. No new failures.
- I ran the 6 new `test_lint.py` tests on this branch: all pass. I checked them out against `main`'s `lint.py`: all 6 fail with `KeyError: 'duplicate_stems'`. They exercise the new code, not something pre-existing.
- I reproduced the PR's own example. Wrote `concepts/vector-search.md` ("The retrieval technique") and `entities/vector-search.md` ("Pinecone (the product)") into a scratch vault, ran `obsidian-wiki lint`:

```
duplicate_stems: 1
  {"stem": "vector-search", "pages": ["concepts/vector-search.md", "entities/vector-search.md"]}
```

- `duplicate_titles` stayed `0` on that same vault. Titles differ, so it misses the collision, same as the PR argues.
- Default `lint` exits 0 (warn). `lint --strict` exits 1. That's the migration behavior the PR describes.

No discrepancies. This works as described.

zorionarrillaga:
Thanks for going through it that carefully — fresh clone, running the six new tests against `main` to watch them fail, reproducing the example and checking the exit codes is more than I expected, and `KeyError: 'duplicate_stems'` on `main` is the right way to prove the tests actually bind.

Your merge settles the `_bootstrap/` question I raised. Following it up properly, the scope difference is wider than I flagged, in **both** directions — `duplicate_stems` borrows `graph_analysis`'s page selection deliberately (so it keys as the merge does), and that selection is not the one the rest of `lint.py` uses:

| directory | `lint.SKIP_DIRS` | `graph_analysis.SKIP_DIRS` |
|---|---|---|
| `_bootstrap`, `.git` | skipped | **not** skipped |
| `_meta`, `_readouts` | **not** skipped | skipped |

So `duplicate_stems` reports collisions under `_bootstrap/` that the other checks ignore — the direction I raised — and it also **misses** collisions under `_meta/` and `_readouts/`, which I did not raise and which is the direction that loses information: `duplicate_titles` reads those pages, `duplicate_stems` does not.

Both follow from the borrowed selection, so it is behaving as written. The gap is only that nothing in the report says this one check has a different scope from its neighbours.

Happy to send a small follow-up: either a docstring/README note stating the scope, or an intersection of the two sets for this check if you would rather it simply match. Your call which — I did not want to pick that one for you twice.

PR #200, opened 2026-08-30
fix: preserve reserved leading-underscore categories in write_page's slug

### LINKED ISSUE #199: `memory_write` with `category: "_raw"` writes to `raw/`, which no skip list excludes
A rough capture written through the documented path ends up packed as compiled knowledge.

## What happens

The MCP tool description tells the caller to use `_raw` (`server.py:160`):

```python
mcp.tool(name="memory_write", description="Write a wiki page. Use category '_raw' for a rough capture.")(write_page)
```

and `docs/deployment.md:74-76` says the same thing for the HTTP route:

> Use `category: "_raw"` for a rough capture you intend to promote with `wiki-ingest` later — the
> same role `_raw/` plays for the `wiki-capture` skill.

`write_page` slugs both segments (`server.py:77`):

```python
rel = f"{_slug(category)}/{_slug(title)}.md"
```

and `_slug` (`server.py:52-53`) replaces every run of characters outside `[a-z0-9]` with `-`, then
strips leading and trailing `-`:

```python
def _slug(title: str) -> str:
    return re.sub(r"[^a-z0-9]+", "-", title.lower()).strip("-") or "untitled"
```

So the leading underscore is dropped and the page lands in `raw/`, not `_raw/`. Each reserved name
I passed behaves the same way:

| `category` | directory written |
|---|---|
| `_raw` | `raw` |
| `_archived` | `archived` |
| `_staging` | `staging` |
| `_archives` | `archives` |
| `_meta` | `meta` |
| `.obsidian` | `obsidian` |

`_raw` is excluded by four skip lists — `graph_analysis.py:73-75`, `lint.py:19`,
`context_pack.py:16`, `trust.py:27-29`. `raw` is in none of them.

## What I measured

At `37596cf`, a vault holding one real page plus one capture written the documented way:

```
write_page("Half formed thought", "_raw", "Rough capture about vector clocks, not reviewed.")
  -> raw/half-formed-thought.md

graph_analysis.iter_pages   : ['concepts/vector-clocks.md', 'raw/half-formed-thought.md']
build_context_pack(topic=…) : 2 pages included
trust.iter_trust_pages      : ['raw/half-formed-thought.md']
lint confidence_missing_fields : [{'page': 'raw/half-formed-thought.md',
                                   'missing': ['base_confidence', 'lifecycle']}]
```

Moving that one file to `_raw/` and changing nothing else:

```
graph_analysis.iter_pages   : ['concepts/vector-clocks.md']
build_context_pack(topic=…) : 1 page included
trust.iter_trust_pages      : []
```

The context-pack line is the one that seems to matter most, because it is the outcome
`docs/configuration.md:141` already warns about in your words — *"makes superseded drafts
retrievable and citable as though they were compiled pages"* — reached here through the documented
capture path rather than through a mis-scoped `qmd collection add`.

## Directions, if it is worth fixing

I did not send a patch because the reasonable fixes point in different directions and the choice
looks like yours:

- preserve a leading `_` on the category segment, so `_raw` stays `_raw`. I tried this one: the
  capture then lands in `_raw/`, which is the second measurement block above. It does not change the
  `.obsidian` row of the table;
- normalise or reject the reserved names specifically — a small blocklist over constants that
  already exist — rather than an allowlist over categories, since `docs/architecture.md:27` says the
  schema is meant to expand as domains are added;
- or leave the behaviour and adjust the documentation, if writing into `_raw/` through the API was
  not the intent. That would be two places: `server.py:160` and `docs/deployment.md:74-76`.

## What I did not check

`.skills/wiki-ingest/SKILL.md:85` defines raw-mode promotion over `OBSIDIAN_VAULT_PATH/_raw/` (or
`OBSIDIAN_RAW_DIR`), but the skills are agent instructions rather than code, so I read them and did
not run them — I do not know what promotion does with a capture sitting in `raw/`. `server.py` was
imported with `fastapi`, `pydantic` and `mcp` stubbed; `write_page` can raise `fastapi.HTTPException`
(`server.py:80`, and via `_resolve` at `server.py:48`), but neither is raised on the path I
exercised. The server itself was not started, and the vault was one I built.

AmirF194:
`write_page`'s rel-path build calls `_slug(category)` directly, and `_slug()` strips
every leading non-alphanumeric character (`re.sub(r"[^a-z0-9]+", "-", ...)`). A page
written with `category="_raw"` therefore lands at `raw/<title>.md`, not `_raw/<title>.md`.

That matters because four separate skip lists (`graph_analysis.SKIP_DIRS`,
`lint.SKIP_DIRS`, `context_pack.SKIP_DIRS`, `trust.TRUST_SKIP_DIRS`) all exclude an
unreviewed capture by an exact match against `"_raw"` (and its siblings, e.g.
`"_archived"`). None of them match `"raw"`, so a page meant to be excluded from graph
analysis, linting, context packing and trust scoring gets pulled into all four as if
it were reviewed knowledge.

Added `_category_slug()`, which slugifies the category and re-adds the leading
underscore when the caller's category started with one, and pointed both of
`write_page`'s uses of `_slug(category)` (the path build and the frontmatter
`category:` field) at it. `_slug(title)` is unchanged; only the category segment is
reserved-name-aware.

**Verified:**
- New regression test (`test_reserved_raw_category_keeps_its_underscore`) fails on
  main (`written["path"] == "raw/clipped-note.md"`) and passes on this branch.
- Full `pytest tests/` green on Python 3.12 with the `[server]` extra installed
  (690 passed, 4 skipped, one pre-existing failure in
  `test_code_understand_codegraph_via_bin_env` unrelated to this change and present
  on main before this diff, a `python3` PATH-resolution artifact of my minimal
  container image, not of the repo).
- Did not verify against Python 3.9: `mcp>=2.0` (the `[server]` extra) has no
  distribution for 3.9, so `tests/test_server.py` is skipped by `importorskip` on
  that interpreter both on main and on this branch, in CI and locally.

Fixes #199

### DIFF
--- a/obsidian_wiki/server.py
+++ b/obsidian_wiki/server.py
@@ -53,6 +53,12 @@ def _slug(title: str) -> str:
     return re.sub(r"[^a-z0-9]+", "-", title.lower()).strip("-") or "untitled"
 
 
+def _category_slug(category: str) -> str:
+    """Slugify a category, preserving a reserved leading underscore."""
+    slug = _slug(category)
+    return f"_{slug}" if category.lower().startswith("_") else slug
+
+
 def search(q: str, limit: int = 8) -> dict[str, Any]:
     return graph_query(VAULT, q, top_n=limit)
 
@@ -74,7 +80,7 @@ def write_page(
     summary: str = "",
     upsert: bool = True,
 ) -> dict[str, Any]:
-    rel = f"{_slug(category)}/{_slug(title)}.md"
+    rel = f"{_category_slug(category)}/{_slug(title)}.md"
     target = _resolve(rel)
     if target.exists() and not upsert:
         raise HTTPException(409, f"page already exists: {rel}")
@@ -88,7 +94,7 @@ def write_page(
         [
             "---",
             f"title: {title}",
-            f"category: {_slug(category)}",
+            f"category: {_category_slug(category)}",
             "tags: [" + ", ".join(tags or []) + "]",
             "sources: [" + ", ".join(sources or []) + "]",
             f"summary: {summary}" if summary else "summary:",
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -84,3 +84,12 @@ def test_write_cannot_escape_the_vault(client, tmp_path):
     # The category is slugified before it becomes a directory, so dots never survive.
     assert ".." not in resp.json()["path"]
     assert not (tmp_path.parent / "escape.md").exists()
+
+
+def test_reserved_raw_category_keeps_its_underscore(client):
+    # The four skip lists exclude a page by exact path match against "_raw";
+    # a dropped leading underscore would land it in "raw/" instead.
+    written = client.post("/v1/pages", json={
+        "title": "Clipped Note", "category": "_raw", "content": "x",
+    }).json()
+    assert written["path"] == "_raw/clipped-note.md"


Ar9av:
Tested on st3ve (fresh clone, PR head `78ac04b`).

- Full suite with the `[server]` extra: 692 passed, 3 skipped, 0 failures. The PR reported "690 passed, 4 skipped, one pre-existing failure"; that failure was a `python3` PATH artifact of the author's container and doesn't show up here.
- Checked out `test_reserved_raw_category_keeps_its_underscore` against `main`'s `server.py`: it fails with `'raw/clipped-note.md' != '_raw/clipped-note.md'`, exactly as claimed. Passes on this branch.
- Went past the string comparison and called `write_page(title, "_raw", ...)` directly against the real skip lists:
  - On `main`: the page lands at `raw/clipped-note.md`, and `graph_analysis.SKIP_DIRS` does not exclude it. The actual bug reproduces — an unreviewed `_raw` capture leaks into graph analysis.
  - On this branch: the page lands at `_raw/clipped-note.md`, correctly excluded by both `graph_analysis.SKIP_DIRS` and `lint.SKIP_DIRS`.

No discrepancies. This fix works as described. Merging.

AmirF194:
Thanks for going past the test assertion and checking it against the real skip lists directly, that is the validation that actually matters here. Glad it holds up.

PR #202, opened 2026-08-31
fix: normalize relative paths in manifest delta matching

Yiji2656:
## Summary

- Normalize mixed `/` and `\` separators before indexing and matching relative manifest paths.
- Apply host-specific path normalization and case handling during comparison.
- Add cross-platform regression tests, including a Windows path with a non-ASCII username.

## Problem

On Windows, manifest entries stored with POSIX-style separators were not matched against scanned paths using backslashes. As a result, previously ingested sources could be reported as `NEW` instead of `MOD` or unchanged.

## Testing

- `python -m pytest tests/test_manifest_delta.py -q`
  - 8 passed
- `python -m pytest tests/test_manifest_delta.py tests/test_cache_manifest_shapes.py tests/test_cache.py -q`
  - 55 passed

### DIFF
--- a/scripts/manifest.py
+++ b/scripts/manifest.py
@@ -112,6 +112,12 @@ def _skip_patterns(cli_skip: str | None) -> list[str]:
     return pats
 
 
+def _normalize_for_match(path: str) -> str:
+    """Normalize either slash style for host-independent path comparison."""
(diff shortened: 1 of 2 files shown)


Ar9av:
Oh, this makes sense
Thanks @Yiji2656

PR #204, opened 2026-08-31
fix(lint): stop flagging embeds and normalise .md/escaped-pipe wikilink targets

### LINKED ISSUE #176: lint_vault reports image/PDF/canvas/.base embeds and table-escaped pipes as broken links

AmirF194:
## Root cause

`broken_links` in `lint_vault` walks every page's `links` list against `by_slug`, an
index built only from `.md` files (`_iter_pages` globs `*.md`). `_parse_page`'s
link-harvest doesn't know that, so it hands the check three kinds of targets `by_slug`
was never going to contain:

- an embed with a non-`.md` extension (`![[diagram.png]]`, `[[board.canvas]]`,
  `[[dashboard.base]]`) names an attachment, not a page
- `[[alpha.md]]`, an explicit `.md` suffix, while `by_slug` keys on the bare stem
- `[[beta\|B]]`, a Markdown-table-escaped pipe: `_WIKILINK_RE`'s own trailing
  `(?:[|#][^\]]*?)?` group consumes the `|B` as an alias, leaving a trailing backslash
  in the captured target

`_normalise_node_id` already strips an explicit `.md` suffix when building a page's
own id, so the harvest side just never reused that normalisation. On a small vault with
any attachments, this makes the lint check permanently fail and buries the one real
broken link in six false positives.

## Fix

`_wikilink_page_target()` returns `None` for a target whose extension isn't `.md`
(skipped, since `by_slug` can never validate it either way), and otherwise strips a
trailing backslash and an explicit `.md` suffix before `_parse_page` slugs it.

## Verification

- New test `test_lint_vault_broken_links_ignores_embeds_and_normalises_md_and_escaped_pipe`
  fails on `main`, passes on this branch (checked both ways in the same container).
- Full suite in `python:3.12-slim` and `python:3.9-slim` (the CI matrix's two legs):
  689/690 pass on 3.12, one pre-existing failure unrelated to this change
  (`test_code_understand_codegraph_via_bin_env`, a `python3`-PATH artifact of the
  minimal image, reproduces identically on `main`); 3.9 leg passes `test_lint.py` clean.
- Did not verify beyond the character classes this fix handles; a target with an
  extension other than `.md` is always skipped rather than checked for existence, which
  matches the issue's own suggested fix and keeps this change to the harvest's
  normalisation rather than adding a new attachment-existence check.

Fixes #176

### DIFF
--- a/obsidian_wiki/lint.py
+++ b/obsidian_wiki/lint.py
@@ -154,16 +154,25 @@ def _normalise_node_id(raw: str) -> str:
     return "/".join(_slug(part) for part in target.strip("/").split("/") if part)
 
 
+# Extensions Obsidian embeds as attachments. Anything else after a dot is part
+# of the page name ("Node.js", "v1.2 release notes"), not a file extension.
+_ATTACHMENT_SUFFIXES = frozenset({
+    ".png", ".jpg", ".jpeg", ".gif", ".bmp", ".svg", ".webp", ".avif",
+    ".mp3", ".wav", ".m4a", ".ogg", ".flac", ".mp4", ".webm", ".mov", ".ogv",
+    ".pdf", ".canvas", ".base",
+})
+
+
 def _wikilink_page_target(raw: str) -> str | None:
-    """Normalise a `_WIKILINK_RE` capture to a page name, or None for a target
-    `by_slug` (`.md`-only) can never hold: an embed's non-`.md` extension, or
-    a table-escaped pipe's trailing backslash.
+    """Normalise a `_WIKILINK_RE` capture to a page name, or None for an
+    attachment embed `by_slug` (`.md`-only) can never hold. Also strips an
+    explicit `.md` suffix and a table-escaped pipe's trailing backslash.
     """
     name = raw.rstrip("\\").split("/")[-1]
     suffix = Path(name).suffix.lower()
     if suffix == ".md":
         return name[: -len(suffix)]
-    if suffix:
+    if suffix in _ATTACHMENT_SUFFIXES:
         return None
     return name
 
--- a/obsidian_wiki/lint.py
+++ b/obsidian_wiki/lint.py
@@ -154,6 +154,20 @@ def _normalise_node_id(raw: str) -> str:
     return "/".join(_slug(part) for part in target.strip("/").split("/") if part)
 
 
+def _wikilink_page_target(raw: str) -> str | None:
+    """Normalise a `_WIKILINK_RE` capture to a page name, or None for a target
+    `by_slug` (`.md`-only) can never hold: an embed's non-`.md` extension, or
+    a table-escaped pipe's trailing backslash.
+    """
+    name = raw.rstrip("\\").split("/")[-1]
+    suffix = Path(name).suffix.lower()
+    if suffix == ".md":
+        return name[: -len(suffix)]
+    if suffix:
+        return None
+    return name
+
+
 def _parse_page(path: Path, vault: Path) -> dict[str, Any]:
     text = path.read_text(encoding="utf-8", errors="replace")
     front_match = _FRONTMATTER_RE.match(text)
@@ -164,7 +178,8 @@ def _parse_page(path: Path, vault: Path) -> dict[str, Any]:
 
     links: list[str] = []
     for raw in _WIKILINK_RE.findall(text):
-        target = _slug(raw.split("/")[-1])
+        name = _wikilink_page_target(raw)
+        target = _slug(name) if name else ""
         if target:
             links.append(target)
     for href in _MD_LINK_RE.findall(text):
--- a/probe.py
+++ b/probe.py
@@ -1,0 +1,19 @@
+import sys, tempfile, pathlib, json
+sys.path.insert(0, ".")
+from obsidian_wiki.lint import lint_vault
+
+def page(v, rel, links=()):
+    p = v / rel; p.parent.mkdir(parents=True, exist_ok=True)
+    body = "\n".join(f"- [[{l}]]" for l in links)
+    p.write_text(
+        "---\ntitle: T\ncategory: concepts\ntags: [a]\nsources: []\n"
+        "created: 2026-01-01\nupdated: 2026-01-01\nsummary: s\n---\n\n" + body + "\n",
+        encoding="utf-8")
+
+v = pathlib.Path(tempfile.mkdtemp()) / "vault"
+page(v, "entities/Node.js.md")
+page(v, "concepts/index.md", links=["Node.js"])
+r = lint_vault(v)
+print("broken_links:", r["findings"]["broken_links"])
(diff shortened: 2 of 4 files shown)


Ar9av:
Good catch on the root cause — the three-way breakdown (attachment embed, explicit `.md`, table-escaped pipe) is right, and the new test genuinely fails on `main`.

One regression, which I pushed a fix for onto this branch (thanks for leaving maintainer edits on).

`_wikilink_page_target` treated **any** suffix as a file extension:

```python
if suffix:
    return None
```

`Path("Node.js").suffix` is `".js"`. Those links then vanish from `links` entirely — and `links` feeds `orphan_pages` and `stats.link_count`, not just `broken_links`. Measured on this branch vs `main`, with `entities/<name>.md` existing and `concepts/index.md` linking to it:

| page name | main | this branch (before 18e6577) |
|---|---|---|
| `Node.js` | link_count=1, orphans=[] | link_count=0, **reported orphan** |
| `Next.js` | link_count=1, orphans=[] | link_count=0, **reported orphan** |
| `v1.2 release notes` | link_count=1, orphans=[] | link_count=0, **reported orphan** |
| `GPT-4.1 eval` | link_count=1, orphans=[] | link_count=0, **reported orphan** |

So it traded false positives in `broken_links` for false positives in `orphan_pages`, on exactly the entity pages a tech wiki tends to have.

`18e6577` narrows the skip to an `_ATTACHMENT_SUFFIXES` allowlist (Obsidian's image/audio/video types plus `.pdf`/`.canvas`/`.base`) and adds `test_lint_vault_keeps_links_to_pages_whose_name_contains_a_dot`, which fails on `db9343c` and passes after. Your `.md`, escaped-pipe, and embed cases are untouched — all 37 `test_lint.py` tests pass.

Shout if you'd rather handle it differently; otherwise this looks ready once CI is green.

AmirF194:
Confirmed: all 37 tests pass at 18e6577, including the new Node.js/Next.js case, and your `_ATTACHMENT_SUFFIXES` allowlist is the right fix, mine was too broad. One thing before merge: `probe.py` and `probe2.py` at the repo root look like scratch repro scripts rather than something meant to ship.

Ar9av:
got it, thanks @AmirF194

PR #207, opened 2026-09-04
fix(skills): parser-safe title/summary in canonical page template and server write_page

bacca87:
## Why

The canonical **Page Template** in `llm-wiki/SKILL.md` — the template every write skill follows (`wiki-ingest`: *"Use the page template from the llm-wiki skill"*) — showed `title:` and `summary:` as **bare YAML scalars**. Free-text values containing `": "` (colon + space), `"#"`, or quotes break YAML parsing: Obsidian then reports **"Invalid properties"** and hides the frontmatter.

This is not hypothetical. A real vault ingesting colon-heavy technical documentation ended up with **25 unparseable pages** from a single ingest pass — summaries like:

```yaml
summary: Inventario delle entità DB del WMS documentate nella guida SDM: inventory (Item/Sku/StockUnit/...), storage (locations/positions/operation points), handling (Task/TO), shipping (Shipment/LoadPlan).
```

The second `: ` makes the whole frontmatter block fail to parse.

## Prior art

- **#3** (merged) fixed this exact footgun, but only in `wiki-update/SKILL.md` (folded scalar `>-` + guidance). The canonical template every *other* write skill follows was left unguarded.
- **#2** (quoted variant, same day) was closed unmerged.
- **#103**'s OKF export now auto-repairs "an unquoted colon in `summary:`" at read time — treatment, not prevention.

This PR fixes the source.

## What's in it

- **`.skills/llm-wiki/SKILL.md`** — the Page Template (and the project-overview example) now emit `title`/`summary` as folded scalars (`>-`), with an explicit **Parser-safe scalars** paragraph matching wiki-update's merged guidance: a bare scalar containing `": "`, `"#"`, or quotes breaks YAML parsing, and the value stays indented under `title: >-` / `summary: >-`.
- **`obsidian_wiki/server.py`** — `write_page` emits `title`/`summary` via `json.dumps`. YAML 1.2 is a superset of JSON, so the emitted double-quoted scalar always parses, for any caller-supplied value (colons, `#`, embedded quotes). Empty `summary` still emits the bare `summary:` key.
- **`tests/test_server.py`** — regression test: writes a page whose title/summary contain `": "`, `"#"`, and quotes, then asserts both scalars round-trip through a JSON parse of the emitted value (a JSON parse of a `json.dumps` scalar proves valid YAML escaping and exact value survival). The existing round-trip assertion is updated to the quoted form.

## Validation

- `pytest tests/test_server.py` — **11 passed** (10 existing + 1 new), including the updated round-trip test.
- Full suite on this Windows machine has pre-existing environment failures (config/XDG tests); the failing-test list is **identical to a pristine `origin/main` worktree** — no regressions introduced.
- Hand-checked: a title/summary with `": "`, `"#"`, `"` emits `title: "Kafka: \"Rebalance\" #notes — offsets: 42"` and parses back to the exact original value.

### DIFF
--- a/.skills/llm-wiki/SKILL.md
+++ b/.skills/llm-wiki/SKILL.md
@@ -88,7 +88,8 @@ Each project directory has an overview page structured like this:
 
 ```markdown
 ---
-title: My Project
+title: >-
+    My Project
 category: project
 tags: [ai, web, backend]
 source_path: ~/.claude/projects/-Users-name-Documents-projects-my-project
@@ -161,7 +162,8 @@ When creating a new wiki page, use this structure:
 
 ```markdown
 ---
-title: Page Title
+title: >-
+    Page Title
 category: concepts
 tags: [ml, architecture]
 aliases: [alternate name]
@@ -169,7 +171,8 @@ relationships:
   - target: "[[concepts/related-concept]]"
     type: extends
 sources: [papers/attention.pdf]
-summary: One or two sentences, ≤200 chars, so a reader (or another skill) can preview this page without opening it.
+summary: >-
+    One or two sentences, ≤200 chars, so a reader (or another skill) can preview this page without opening it.
 provenance:
   extracted: 0.72
   inferred: 0.25
@@ -202,6 +205,8 @@ Things that are unresolved or need more sources.
 
 - [[references/attention-is-all-you-need]] — Original paper
 ```
+
+**Parser-safe scalars.** Write free-text frontmatter values — at minimum `title` and `summary` — with folded scalar syntax (`>-`) as shown above: a bare scalar containing `: ` (colon + space), `#`, or quotes breaks YAML parsing, and Obsidian then reports "Invalid properties" and hides the frontmatter. Keep the value indented on the line(s) following `title: >-` / `summary: >-`.
 
 ## Paper Deep-Dive Template
 
--- a/obsidian_wiki/server.py
+++ b/obsidian_wiki/server.py
@@ -11,7 +11,6 @@ Run it with ``python -m obsidian_wiki.server``.
 from __future__ import annotations
 
 import hmac
-import json
 import os
 import re
 from contextlib import asynccontextmanager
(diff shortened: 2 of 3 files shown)


Ar9av:
Confirmed the bug and the prior art. `write_page` really did emit bare scalars, the `.skills/llm-wiki` template really was the unguarded one every other write skill follows, and #3 only ever covered `wiki-update`. Reproduced your 11-passing `tests/test_server.py` in a venv with the `server` extra.

The SKILL.md half is right as written. The `server.py` half needed a change, pushed as `55cab0c`.

`json.dumps` does satisfy a YAML 1.2 parser — but this vault has a second reader, and it is not a YAML parser. `lint._parse_frontmatter_values` ends with:

```python
values[key] = value.strip("'\"")
```

It strips the outer quotes and never unescapes. That is what feeds `broken_links`, orphan detection, the missing-summary check, and `wiki-query`'s index-first pass over `summary:`. Emitting JSON hands it escapes it cannot read:

| value | pyyaml | `_parse_frontmatter_values` |
|---|---|---|
| `Kafka: "Rebalance" #notes offsets: 42` | ok | `Kafka: \"Rebalance\" #notes offsets: 42` |
| `Inventario delle entità DB: guida SDM` | ok | `Inventario delle entità DB: guida SDM` |
| `向量时钟: 分布式系统` | ok | `向量时钟: ...` |
| `He said "hi"` | ok | `He said \"hi\` — `strip()` ate a backslash |

The `ensure_ascii=True` rows are also what lands **on disk**, not just what the parser sees — the motivating vault in your description is Italian, so `entità` would be written into the markdown file as `entità` for the human who opens it.

Folded blocks pass both readers on every case above, and `_parse_frontmatter_values` has understood `>-` since it was written (`if value in {">", ">-", "|", "|-"}`). `55cab0c` swaps `json.dumps` for a `_folded()` helper, so the server emits the same convention this PR documents:

```yaml
title: >-
  Inventario delle entità DB: guida SDM — perché
summary: >-
  Kafka: "Rebalance" #notes
```

`pytest tests/test_server.py` — 11 passed. I kept your regression test's intent and strengthened it: it now asserts the folded form on disk **and** that `_parse_frontmatter_values` returns the exact original value, which is the assertion that would have caught this.

One deliberate behavior note: `_folded` collapses internal whitespace (`" ".join(value.split())`), since a folded block cannot carry a raw newline and the read side joins folded lines with spaces anyway. For single-line `title`/`summary` that is a no-op.

Nice find — the #3/#2/#103 archaeology made this easy to review.

Issue #208, opened 2026-09-08
Feature: Add a local single-vault Web Console for browsing and managing Obsidian-Wiki


lordraiden:
## Summary

Add an optional local web administration console to `obsidian-wiki` that makes an agent-maintained vault easier to inspect, review, and operate from a browser.

The first version should be a **single-user, single-vault local console**. It should not replace Obsidian, become a multi-user SaaS, or orchestrate LLM agents. Its purpose is to provide an operational control surface for the knowledge pipeline: agent-generated changes, source freshness, staged review, vault health, and controlled maintenance actions.

The vault must remain the source of truth: plain Obsidian-compatible Markdown files, frontmatter, wikilinks, `.manifest.json`, and the existing project conventions.

## Problem

`obsidian-wiki` is intentionally agent-first:

- An agent ingests, distills, merges, and maintains knowledge.
- The vault stores the durable output as Markdown.
- Obsidian is the primary visual viewer and editor.
- The CLI and skills expose deterministic maintenance and agent-driven workflows.

This is powerful, but routine ownership and operational tasks are fragmented for users who want a quick browser-based control surface:

- See which sources were ingested, which changed, and which are still pending.
- See what agents changed recently and what is awaiting human review.
- Inspect broken links, orphan pages, duplicates, contradictory content, and other health signals.
- Review staged AI-generated changes before promoting them.
- Run controlled maintenance actions and inspect their results without depending on terminal output.
- Check whether the vault is clean, synchronized, and ready for an agent to use.

Obsidian already provides excellent page browsing, full-text search, Markdown editing, backlinks, graph visualization, and note authoring. Duplicating those capabilities in a browser UI would create a weaker second editor and dilute the project’s focus.

The web console should therefore complement Obsidian rather than compete with it: **Obsidian remains the place to read, navigate, write, and visualize notes; the web console becomes the place to operate and govern the agent-maintained knowledge pipeline.**

### Phase 1 — Local Vault Operations Viewer

Estimated effort: **3–5 engineering days** for one developer familiar with the codebase.

This phase is read-only. It should be useful even for users who want to continue doing all note reading and editing in Obsidian.

#### User-facing capabilities

1. **Operations dashboard**
   - Total page and category counts as high-level operational context.
   - Recently changed pages, with path, timestamp, source, and whether the change is staged or live.
   - Counts for `_raw/` and `_staging/`.
   - Manifest/source freshness summary: new, modified, unchanged, and pending sources.
[...]

PR #210, opened 2026-09-09
feat(server): read-only operations console at /ui

Ar9av:
Addresses the Phase 1 scope of #208 — a browser view of vault operations.

## What's here

| Route | Returns |
|---|---|
| `GET /ui` | Static console shell — no build step, no dependencies, no vault data |
| `GET /v1/status` | Page/category counts, `_raw`/`_staging` depth, manifest source freshness, recent changes, git status |
| `GET /v1/health` | `run_doctor()` + `lint_vault()` |
| `GET /v1/staging` | Staged pages with a unified diff against the live page |

Everything is derived from the vault at request time — no database, no cached state. The health route calls the same Python the CLI calls rather than shelling out and parsing terminal output.

## Deliberately not here

**Staged promote/discard.** #208 asks to "reuse the existing staged-write semantics and `/wiki-stage-commit` behavior." That behavior lives in `.skills/wiki-stage-commit/SKILL.md` as agent instructions — `grep -rn "_staging" obsidian_wiki/` returns nothing. A promote button would mean writing a second promotion workflow, so this view links out to the skill instead. Porting that logic to Python is worth its own issue.

**Session cookies, CSRF, a login flow.** The shell is static and carries no vault data, and fetches with the same bearer key API clients already use, held in `sessionStorage`. No cookie means no CSRF surface.

**A sanitiser dependency.** `difflib` emits plain text; the client inserts it with `textContent`, which never invokes the HTML parser.

## Tests

Six new tests in `tests/test_server.py`: key enforcement on all three routes, counts derived from a fixture vault (including root-level `index.md` not being treated as a category), new-vs-update staging diffs, doctor+lint shape, and a shell that renders no vault content.

The XSS test stages a page containing `<script>` and an `onerror` handler and asserts no element is constructed from it.


Ar9av:
Verified on a Linux box against a live server, not just CI.

- Fresh clone of the branch, `pip install -e ".[server]"`, **17/17 tests pass** on Python 3.12
- Ran the server against a real 72-page vault and a synthetic one; `/ui`, `/v1/status`, `/v1/health`, `/v1/staging` all serve correctly
- All three `.manifest.json` container shapes read (`sources`, `projects`, `research_sessions`), and a `pages_in_vault` entry deleted from disk is correctly flagged under **missing**
- `git` panel reports branch and uncommitted count off a real repo
- XSS probe: a staged page containing `<script>` and an `onerror` handler renders as diff text — `imgs: 0, scripts: 0`, `document.title` unchanged
- `/ui` shell contains no vault content (grep for staged page names returns 0)

One bug found and fixed during this run (c9f74bc): the first version only read `manifest["sources"]`, so the sources panel was silently empty on any vault built by `wiki-update` or `wiki-research`.

Screenshots are from a synthetic vault built for this run.

dashboard

sources and staging

Issue #216, opened 2026-09-11
Proposal: machine-portable source keys for a multi-machine vault


JaySon-Huang:
Thanks for sharing obsidian-wiki. I have been running my own vault on it, and it has made that wiki noticeably easier to keep in order than the setup I was using before. This issue describes one problem I ran into while using it across two machines — and a direction I would like your thoughts on.

## How I use the wiki

I run one vault across two machines, synced over git:

- **A capture machine**, connected to my IM. Through an agent it adds and updates the articles, links and notes I come across during the day — that is its main job.
- **My MacBook**, my main machine, where I do code development and deeper research.

Both check out the same vault. Git syncs them, and I let an LLM agent handle merges. Writes rarely collide — the two machines tend to touch different material — so conflict resolution is not the pain point. Reading is. Whichever machine I am on, I want to query and synthesize knowledge that was captured on the *other* one, and that has to work without the other machine being present.

## What breaks when I read

A source that the capture machine ingested is recorded in `.manifest.json` under a path on that machine (or under its `$HOME`). On the MacBook that path means nothing, so:

- **The identity does not travel.** The same article gets a different key depending on which machine ingested it, and a key written on one machine cannot be matched on the other. That is the same identity split the framework's keyed-by-raw-string design is meant to prevent — it just happens through the machine path rather than through `~` vs absolute spelling.
- **The provenance does not travel.** A page's `sources:` frontmatter holds the same machine path, so the "where did this come from" trail is broken on the machine I am actually reading from.
- **A source from the other machine looks lost, and the tooling can act on that.** The cache layer cannot tell "this lives on the other machine" from "this file is gone here", so it reports `missing` — and `wiki-ingest` documents `missing` as "skip and optionally clean up". A read on one machine can therefore quietly drop an entry whose file is perfectly fine on the other.

## Evidence from my vault

From my own vault after roughly two months of this setup. Read-only, reproducible:

**Source keys were keyed by machine path.** At the commit before I migrated it, the manifest held 50 sources:

| key form | count |
|---|---|
| absolute machine path | **26** |
| pseudo-key (bare `https://…`, plus one custom scheme) | 21 |
| vault-relative | 3 |

25 of the 26 absolute keys were rooted at that machine's vault checkout; the 26th pointed into another host's agent cache (`/home/hermes_agent/.hermes/cache/...`). Note the mix: a single directory (`Raw/database/`) held both absolute-keyed and relative-keyed sources.

```bash
# read-only
[...]

Issue #88, opened 2026-06-11
Manifest updates are ad-hoc inline Python; add a stdlib helper


Ar9av:
## Problem

The ingest skills say 'update `.manifest.json`' but provide no helper. In practice the agent writes raw inline Python each run to read/merge/write the manifest — verbose, easy to get quoting wrong, and it doesn't normalize paths (see #PATHS) or honor skip config (see #SKIP).

## Fix
Add `scripts/manifest.py` (pure stdlib, no deps, matches the existing `scripts/` convention):

- `normalize <vault>` — rewrite source keys to canonical absolute paths, merge collisions (keep newest), `--dry-run` to preview.
- `delta <vault> --scan <glob> [--skip a,b]` — list new/modified sources vs manifest, honoring `WIKI_SKIP_PROJECTS` + `--skip`.

Instructions still work without it (it's an optional accelerator), but it makes the manifest steps deterministic and testable.

Issue #86, opened 2026-06-11
Manifest mixes absolute and ~-relative source paths → silent re-ingestion


Ar9av:
## Problem

The manifest stores source keys in two inconsistent forms. The schema example in `wiki-status/SKILL.md` itself documents both:

```json
"sources": {
  "/absolute/path/to/file.md": { ... },
  "~/.claude/projects/-Users-name-my-app/abc123.jsonl": { ... }
}
```

Because the manifest is keyed by the raw string, the **same file** can be tracked under both `~/.claude/...` and `/Users/me/.claude/...`. The delta check in append mode then treats an already-ingested file as new and re-ingests it.

## Evidence
On a real 137-source vault, **5 files were tracked under both forms** — re-ingested redundantly on the next run because the lookup missed the other-form key.

## Fix
- Declare a single canonical form: **source keys are always stored as absolute paths with `~` and env vars expanded.**
- Fix the schema example in `wiki-status/SKILL.md` to use absolute paths consistently.
- Add a normalization + de-dup rule to `llm-wiki/SKILL.md` (merge colliding entries, keep newest `ingested_at`).
- Provide a helper so delta checks and manifest writes normalize before compare/store.

PR #217, opened 2026-09-11
Make manifest source keys machine-portable

JaySon-Huang:
Thanks for obsidian-wiki — I run my vault on it, and it has made the wiki noticeably easier to manage. This PR implements the proposal in the accompanying issue #216: make the persisted source key machine-independent, and keep reading the absolute keys already in the wild.

### What is in this PR

Five commits, one clear boundary each:

1. the contract text,
2. the cache resolution and cross-machine classification,
3. the manifest helper, key-aware `delta`, and `migrate`,
4. the read-only lint finding,
5. the skill and user documentation.

The suite passes locally (771 passed, 4 skipped). I kept it as a single PR rather than splitting the code from the docs, so the whole change can be read in one place.

## Prior art

The absolute-path rule was introduced deliberately to stop a real bug: with the manifest keyed by raw string, the same file tracked as both `~/.claude/…` and `/Users/me/.claude/…` was re-ingested (#86). That diagnosis is right and I am not proposing to undo the intent — one canonical form, never mixed. My case is the same failure class reached from a different direction: the canonical form is itself machine-dependent, so across machines the identity splits again, and pages lose their provenance. This proposal keeps the rule and makes the form portable. (`scripts/manifest.py`'s `normalize`, from #88, is kept as an alias.)

## What I would like your input on

The branch is complete, but the direction is the part worth settling first — before anyone spends time reading the whole diff:

1. **Is a machine-independent canonical key the right direction?** I am not attached to the exact three forms. If you would rather have a single pseudo-key form with a content hash, or something else entirely, that is fine by me — the property that matters is that the persisted identity does not encode a machine, and that reads tolerate the absolute keys already in the wild.
2. **Is `unavailable` the right shape for `cache-check`?** It adds a fourth bucket and narrows `missing` to vault-local sources. That is additive, but it changes the output of an existing command, so I would rather have your opinion than rely on my own judgement.
3. **Is any of the scope better left out?** `migrate`, the read-only lint finding, and the project-identity change are each separable if you would rather take them in stages.

If the documentation half turns out to be the contentious part, the code stands on its own: multi-form key resolution, the `unavailable` bucket and `migrate` are purely additive — they make the mixed reality that already exists in the wild safe, without asking anyone to change what they write.

## What it looks like on my vault

To check the effect rather than only describe it, I made a worktree of my vault at the commit *before* I migrated it by hand — 50 sources, **26 of them keyed by absolute machine paths** — and ran this branch's tooling on that copy.

The cross-machine case first: the worktree sits at a different path from the machine that wrote the keys, so nothing can be stripped lexically. The command says so rather than claiming success:

```console
$ python3 scripts/manifest.py migrate <vault> --dry-run
  WARN   no portable form (kept absolute): /DATA/disk1/…/Clippings/DUCKDB…（上）.md
  … (25 more)
  HINT   26 absolute key(s) could not be made portable. If this vault moved
         machines, pass the old vault root with --from-root <old-vault-root> …
sources: 50 -> 50 (0 re-keyed, 0 collisions merged, 26 kept non-portable)
```

Supplying the root it actually came from:

```console
$ python3 scripts/manifest.py migrate <vault> --from-root /DATA/disk1/…/oh-my-wiki
  WARN   no portable form (kept absolute): /home/hermes_agent/.hermes/cache/documents/…SyncPoint….md
sources: 50 -> 50 (25 re-keyed, 0 collisions merged, 1 kept non-portable)
wrote <vault>/.manifest.json

$ # second run: idempotent, and still honest about the one key that is not portable
sources: 50 -> 50 (0 re-keyed, 0 collisions merged, 1 kept non-portable)
nothing portable to write — 1 key(s) kept non-portable
```

The manifest is now 28 vault-relative + 21 pseudo-keys + 1 absolute. The one that stays absolute is a document living on a *different* host (`/home/hermes_agent/.hermes/cache/...`): it has no portable path form, so the tool keeps it, warns, and points at an explicit key instead of inventing one. That last part is a real gap worth flagging: `cache-update --key` hashes the source file, so it cannot re-key an entry whose file is not on this machine. In my manual migration I renamed that key to an `external/…` prefix by hand; a `migrate --rekey` (or similar) would be a cleaner answer.

**The identity resolves again.** Scanning the vault's own source directories, before and after:

| directory | before migration | after migration |
|---|---|---|
| `Raw/` | 20 new | 4 new, 16 recognized |
| `Clippings/` | 7 new | 2 new, 5 recognized |
| `Designs/` | 1 new | 1 recognized |

Before, every in-vault source looked *new* on this machine, because its stored key named a path from another machine. After, the 22 already-ingested sources are recognized; the 6 that remain new are files genuinely absent from the manifest (later clippings). The `modified` readings are not real edits — a fresh `git checkout` rewrites mtimes, and `content_hash` is what actually prevents redundant work.

And the cache layer no longer calls the other machine's source lost:

```console
missing     : 0
unavailable : /home/hermes_agent/.hermes/cache/documents/…SyncPoint….md
```


Ar9av:
Thanks for this, JaySon. The writeup in #216 did most of the work for me: the diagnosis, the evidence from your own vault, and the alternatives you already ruled out. I ran the branch on a clean Ubuntu box against a synthetic two-machine vault rather than reading the diff, and it holds up.

### Answers to your three questions

- **Machine-independent key: yes, take it.** The `~`-relative rung earns its place; it keeps the file openable, which a content hash would not. I have no appetite for a fourth form.
- **`unavailable`: yes, keep it.** A synced vault reporting another host's source as `missing` is the part that could destroy data, since `wiki-ingest` treats `missing` as cleanup-eligible. One extra bucket is cheap next to that.
- **Scope: keep it whole.** `migrate`, the lint finding and the project-identity change all read on the same contract. Splitting them would leave the docs describing a form the code does not write.

### What I verified

- **Suite:** 771 passed, 4 skipped, matching your number.
- **`cache-check` buckets:** a vault-relative source I deleted came back `missing`; the other host's `/home/hermes_agent/...` key and an old-vault-rooted key came back `unavailable`. That split is the one I cared about most.
- **`migrate` without `--from-root`:** warned per key, printed the `--from-root` hint, and reported `26 -> kept non-portable` style output rather than claiming success.
- **`migrate --from-root <old root>`:** re-keyed the old-vault entries to vault-relative, left the foreign-host key absolute with a warning, and a second run wrote nothing.
- **`~`-relative:** a `$HOME` key survived the round trip as `~/.claude/...` and matched an absolute query afterwards.
- **`cache-update`:** an in-vault source stored `Raw/c.md`; one under `/opt` stored the absolute path and warned on stderr; `--key repo:github.com/o/n` re-keyed the entry in place.
- **Lint:** flagged the `/DATA/...` entry in a page's `sources:` and left the vault-relative sibling alone. Read-only, as documented.
- **`normalize` alias:** still works.

### One bug I hit, and pushed a fix for

Your collision merge keeps the first entry it saw, not the newest. Two keys pointing at the same file:

```
"/vault/Raw/a.md": {"content_hash": "old", "last_ingested": "2026-01-01T00:00:00Z"}
"Raw/a.md":        {"content_hash": "new", "last_ingested": "2026-06-01T00:00:00Z"}
```

merged to the January entry. `_newest` compares `ingested_at`, but `cache.py` writes `last_ingested`, so both sides read `""` and the tiebreak never fires. `cmd_delta` reads the same field, which is why every already-ingested file in my run came back `MOD` — the mtime comparison was running against epoch 0, not your git-checkout mtimes.

Pre-existing, not something you introduced. But `migrate --from-root` is what makes it reachable at scale: re-keying a vault's worth of absolute keys into vault-relative ones is exactly where a collision with an existing relative key shows up, and #216 notes your `Raw/database/` already held both forms. A merge that keeps the stale hash costs a re-ingest.

I pushed `fix(manifest): read the ingest timestamp under either field name` to your branch — one helper reading both names, used at both call sites, plus a regression test with the field name `cache.py` actually writes. Collision now keeps June, delta reports 0 modified, suite at 772 passed.

### The `--rekey` gap

You are right that `cache-update --key` cannot re-key an entry whose file lives on another host, since it hashes the source. Worth a follow-up issue rather than growing this PR. `migrate --rekey old=new` reads well to me.

I will merge after a read of the skill docs. Rebase on `main` when you get a chance — #215 landed since you branched.

JaySon-Huang:
Thanks for the careful review — and for testing it on a clean box rather than reading the diff. You're right on the collision bug, and I should have caught it: the `ingested_at` / `last_ingested` drift was already in my notes as an open question, and I never checked whether it had a live consequence. I kept your commit as its own commit, with your authorship.

Rebased onto the current `main` (#215 and #214 both landed since); no conflicts, six commits now.

Re-verified after the rebase: 797 passed, 4 skipped (up from 772 — your two PRs added tests), and the two-machine vault flow is unchanged end to end — `25 re-keyed / 1 kept non-portable`, second run writes nothing, `missing: 0` / `unavailable: 1`.

Happy for you to read the skill docs when you get a moment.

Ar9av:
Looks good, quite some big change

Thanks

PR #221, opened 2026-09-14
fix(cli): resolve cache source args against the vault root as fallback

JaySon-Huang:
## Summary

`cache-check` and `cache-update` resolved relative source arguments against the
shell's CWD only. A vault-relative argument — the same form the manifest itself
stores — only worked when the shell happened to sit at the vault root. This PR
adds the vault root as a fallback resolution root for both commands, and turns
`cache-update`'s missing-source crash into a clean, actionable error.

## Problem

The manifest stores portable, vault-relative source keys (`Raw/database/x.pdf`),
and the library layer reads them that way: `cache.resolve_key` resolves any
relative key against the vault root (source-key contract v2). The CLI layer,
however, resolved its `source` argument against the CWD:

```python
source = Path(args.source).expanduser().resolve()   # CWD-relative
```

So the documented habit of passing a vault-relative path
(`obsidian-wiki cache-update <vault> "Raw/database/x.pdf"`) broke in three ways
whenever the CWD was not the vault root:

1. **Crash** — the CWD-joined path does not exist; `compute_hash` raised a raw
   `FileNotFoundError` traceback.
2. **Silent corruption (worse)** — a same-named path under the CWD got hashed
   instead. `stored_key()` then derived a key outside the vault, which could
   match no existing entry: a spurious manifest entry was appended while the
   real entry kept its stale hash, so subsequent `cache-check` runs reported
   the real source as `unchanged`.
3. **Confusing ergonomics** — the *stored* key is always normalized to
   vault-relative, so absolute paths were always safe; only the resolution step
   disagreed with the storage step about what "relative" means.

## Fix

- New helpers in `cli.py`, used by both `cmd_cache_check` and `cmd_cache_update`:
  - `_source_candidates(vault, raw)` — absolute args keep only themselves;
    relative args resolve against the CWD **first** (long-standing behavior,
    keeps out-of-vault sources working), then against the vault root.
  - `_resolve_source_arg(vault, raw)` — first candidate that exists on disk;
    when none exists, returns the CWD-relative form so read-only callers can
    classify it as missing.
- `cmd_cache_update` treats a missing result as an error: it prints
  `error: source X does not exist (tried <cwd path> and <vault path>)` and
  exits 1, replacing the traceback.
- `cmd_cache_check` keeps its non-fatal semantics: a query path that exists
  nowhere lands in the `missing` bucket, as before.

CWD-first ordering preserves backward compatibility: out-of-vault relative
paths (e.g. ingesting `../notes/x.pdf` from a sibling directory) resolve
exactly as before, even when the vault contains a same-named file.

| Argument form | Before | After |
|---|---|---|
| Absolute path | resolved as-is | unchanged |
| Relative, exists under CWD | CWD-relative | unchanged (CWD wins) |
| Relative, exists only under vault root | crash / wrong file | resolved against vault root |
| Relative, exists nowhere | `FileNotFoundError` traceback | clean error naming both candidates (`cache-update`); `missing` bucket (`cache-check`) |

## Testing

- New `TestCacheCLIVaultRelativeSources` in `tests/test_cache.py` covers:
  vault-relative update from a foreign CWD (stored key stays
  `Raw/database/x.pdf`); vault-relative check from a foreign CWD reporting
  `unchanged`; CWD-relative precedence when both candidates exist; clean
  no-traceback errors for missing relative and absolute sources.
- Full suite: `844 passed, 4 skipped, 19 subtests passed`.

## Notes

- Docs deliberately unchanged: existing examples in `docs/cli.md` already use
  absolute paths, and the storage side (portable key normalization) is
  untouched.
- `cache-hash` is untouched — its `path` argument has no vault relationship.


Ar9av:
Works. Verified on a clean box against a vault + foreign CWD.

**Confirmed:**

- **Vault-relative from a foreign CWD** — `cache-update <vault> Raw/database/x.pdf` crashed with a raw `FileNotFoundError` before, now hashes the vault file and stores `Raw/database/x.pdf`. `cache-check` went from `missing: [<cwd path>]` to `unchanged: [<vault path>]`.
- **Exit codes** — `cache-update` on a missing relative source exits 1 with `error: source Raw/nope.pdf does not exist (tried <cwd> and <vault>)`, same for a missing absolute path. `cache-check` stays non-fatal at 0.
- **Absolute and `~`-relative args** — unchanged, both still normalize to the portable stored key.
- **CWD-first precedence** — with `Raw/database/x.pdf` present in both places, the CWD copy is the one hashed. Backward compatibility holds.
- **Tests** — your three new cases fail against the parent commit and pass on the branch. Suite 844 passed, 4 skipped.

**One thing worth considering, not a blocker.** In the both-exist case the user typed a string that is byte-for-byte a manifest key, and got a second entry under a different key pointing at a different file:

```console
$ cd ~/scratch   # also has Raw/database/x.pdf
$ obsidian-wiki cache-update <vault> Raw/database/x.pdf
{"key": "~/scratch/Raw/database/x.pdf", ...}   # vault's Raw/database/x.pdf still stale
```

That is the failure you describe as problem #2, now intentional rather than accidental, and I agree CWD-first is the right default — a shell argument resolving against the shell is what every other CLI does. But the collision is invisible. A stderr line when more than one candidate exists would cost two lines and make it a choice rather than a surprise:

```
note: Raw/database/x.pdf matched both <cwd path> and <vault path>; using the CWD one
```

Happy to merge without it if you would rather keep the output clean.

Thanks for tracking this one down to the layer disagreement — the CLI and the library having different ideas of what "relative" means is the kind of thing that hides for a long time.

JaySon-Huang:
Thanks for the review and for verifying it end to end. I added the stderr note.
```bash
$ obsidian-wiki cache-update <vault> Raw/database/x.pdf   # both files exist
note: Raw/database/x.pdf matched both <cwd path> and <vault path>; using <cwd path>
```

Rebased onto the latest main, no conflicts. PTAL again when you're free.

Ar9av:
The note is exactly right. Re-tested the branch at 6df9e57.

The note lands on stderr, so the JSON on stdout stays parseable:

```console
$ obsidian-wiki cache-update <vault> Raw/database/x.pdf 2>/dev/null
{"path": ".../pr221b-cwd/Raw/database/x.pdf", "key": "~/pr221b-cwd/Raw/database/x.pdf", ...}

$ obsidian-wiki cache-update <vault> Raw/database/x.pdf 1>/dev/null
note: Raw/database/x.pdf matched both <cwd path> and <vault path>; using <cwd path>
```

Checked it stays quiet in every unambiguous case: vault copy only, CWD copy only, an absolute argument, and CWD sitting at the vault root — that last one is the subtle one, since `_source_candidates` dedupes to a single entry and the `len(candidates) != 2` guard catches it. No false positives.

The rest still holds after the rebase: vault-relative from a foreign CWD resolves to the vault file and stores `Raw/database/x.pdf`; `cache-check` reports `unchanged`; a missing source still names both candidates and exits 1. Four of the five new tests fail against the parent commit. Suite 848 passed, 4 skipped.

Naming the chosen path in the note rather than just flagging the collision is better than what I suggested — it tells you which file you got, not only that there was a choice.

Merging. Thanks for the fix and for the quick turnaround on the follow-up.

PR #224, opened 2026-09-15
docs(wiki-ingest): use `import pymupdf` in the paper-figure recipe

JaySon-Huang:
## What this changes

The "Academic papers" section of `wiki-ingest` told the reader to reach for PyMuPDF's legacy `fitz` module name — once as the named import, and once in the vector-render call as `fitz.Matrix(4, 4)`. Both now say `pymupdf`. Nothing else in the recipe moved: the API sequence (`get_image_info(xrefs=True)` → locate the caption → `extract_image` → `get_drawings` → `get_pixmap`) is untouched.

## Why

PyMuPDF 1.28.2 (released 2026-08-06) added an import-time warning for the legacy module. Its changelog entry reads "Output warning when legacy `fitz` module is imported", and a clean-venv install of 1.28.2 prints this on `import fitz`:

```
warning: The `fitz` API is deprecated and will be removed in future. Use `import pymupdf` instead.
```

The part worth flagging is where that line lands: **stdout, not stderr** (ref https://github.com/pymupdf/PyMuPDF/issues/5100). Discarding stderr leaves the warning in place; discarding stdout removes it. A step in this recipe that has an agent extract a figure and print JSON or a file path therefore comes back with an unparseable first line. Deprecation noise that a script can filter is a nuisance — this one lands in the channel scripts actually parse.

`fitz` is not broken today, and this change is not a fix for a failure. It is a shim module that re-exports the same objects: `fitz is pymupdf` is `False`, but `fitz.Matrix is pymupdf.Matrix` and `fitz.Document is pymupdf.Document` are both `True`, and `isinstance` agrees across the two namespaces. Switching the documented name costs nothing functionally and removes both the stdout pollution and the exposure to the announced removal.

## Verification

Ran the recipe end to end against a real arXiv PDF (1706.03762) under both import names on pymupdf 1.28.2. Both paths located the same figure on page 2 (xref 128, bbox `[196.6, 72.0, 415.4, 394.4]`, 10.3pt above its "Figure 1" caption) and extracted an identical 121244-byte PNG. The only difference in the output was the warning line prefixing the `fitz` run's JSON — which is the failure mode being removed.

Checked that no other artifact in the repo names the old module: `git grep` finds `fitz` only in this one line, and `PyMuPDF`/`pymupdf` only here and as a prose reference in `llm-wiki/SKILL.md`, which cites the recipe by name without naming an import.

`uv run --frozen pytest tests/test_writing_profile_contract.py -q` → 6 passed. No test asserts on this prose.

## Notes for review

I deliberately kept the explanatory parenthetical to one clause — "the `fitz` alias is deprecated". A longer version documenting the minimum version for the `pymupdf` name (1.24.3, from May 2024) was drafted and dropped, since it costs every future skill load to serve a two-year-old version. That clause is still doing work, though: `fitz` is widespread enough in examples and training data that a bare `` `import pymupdf` `` invites a later contributor to "fix" it back. Happy to cut it entirely if reviewers would rather have the shortest possible line.


Ar9av:
Merging this. The stdout detail is the part that makes it worth doing, and it checks out.

On a clean venv, pymupdf 1.28.2:

```console
$ python -c "import fitz" 2>/dev/null
warning: The `fitz` API is deprecated and will be removed in future. Use `import pymupdf` instead.

$ python -c "import fitz" 1>/dev/null
                                        # nothing
```

So it is stdout, and the usual `2>/dev/null` reflex does not touch it. A step that prints JSON comes back unparseable:

```console
$ python -c "import fitz, json; print(json.dumps({'ok': True}))" 2>/dev/null | jq .
json.decoder.JSONDecodeError: Expecting value: line 1 column 1
```

Ran the recipe against 1706.03762 under both names and got your numbers to the byte — xref 128, bbox `[196.6, 72.0, 415.4, 394.4]`, 10.3pt above the caption, a 121244-byte PNG with an identical sha256. The only difference between the two runs was the warning line on the `fitz` one. The vector branch renders fine under `pymupdf.Matrix(4, 4)` too (876x1290 from that clip).

Your aliasing note holds: `fitz is pymupdf` is False, `fitz.Matrix is pymupdf.Matrix` and `fitz.Document is pymupdf.Document` are both True, and `isinstance` agrees across namespaces. `git grep fitz` on main finds the one line you changed, and `llm-wiki/SKILL.md:251` cites the recipe without naming an import.

Keep the parenthetical. You are right that a bare `import pymupdf` invites someone to "fix" it back to `fitz` — the clause is what stops that, and one clause is the right size for it.

Thanks for chasing which channel the warning lands in. A one-word docs change is easy to wave through without that, and it would have been the wrong reason to merge it.

PR #232, opened 2026-09-21
feat(memory): one locked writer for the vault memory surface

Ar9av:
## What this changes

`index.md`, `log.md`, and `hot.md` were maintained by **prose**. Fifteen-odd skills each restated "append a line to the log, add the new pages to the index, rewrite the hot cache" in their own words, rewrote the files wholesale, and took no lock.

Two consequences, both real:

- **Parallel writers dropped updates.** A `batch-plan` ingest fan-out, or the Dockerized server writing a vault a local skill also has open, is a plain read-modify-write race. Whichever write landed second won, and the first vanished with no error.
- **Documented rules were not rules.** The `~500-word` cap on `hot.md` appeared in five places and was enforced in none.

This replaces that with one code path, serialised by the advisory lock the manifest writer already used and written atomically via `os.replace`.

before and after the memory writer

The lock was refactored out of `cache.py` into a reusable `advisory_lock` rather than duplicated; `manifest_lock` is now a thin wrapper over it, and `ManifestLockTimeout` still exists as a subclass so nothing downstream breaks.

## The command surface

`memory sync` is the one call a skill usually wants. It appends the log line, reconciles the index against the pages actually on disk, and regenerates the hot cache — **all under one lock**, so another writer cannot interleave between the three and leave a snapshot describing a vault state that never existed.

the memory commands in a terminal

| Command | What it does |
|---|---|
| `memory status` | Index drift, log size, hot-cache budget, profile and todo counts |
| `memory sync` | Log, index, and hot cache as one locked update |
| `memory log VERB` | Append one parseable operation line |
| `memory index` | Reconcile `index.md` against disk |
| `memory hot` | Regenerate `hot.md` within its word cap |
| `memory recap` | Print the injectable profile + threads + activity block |
| `memory profile list\|set\|forget` | Durable facts about the vault owner |
| `memory todo list\|add\|done\|drop\|prune` | Threads carried between sessions |

## What is generated and what stays yours

This is the distinction reviewers should push on hardest, because getting it wrong would destroy user content.

**`index.md` is reconciled, not overwritten.** One section per category is regenerated from disk; the preamble and any section whose heading is not a category survive verbatim. A hand-written "Reading queue" stays, a stale entry for a deleted page goes.

**`hot.md` is generated except for `## Key Takeaways`.** That is the one slot a model writes on purpose, and it carries across every rebuild unless `--takeaways` replaces it. A fully mechanical hot cache would have thrown away the semantic snapshot that made the file useful.

the generated hot cache

## Two new artifacts

`_meta/profile.md` and `_meta/todos.md` — durable facts about the owner, and threads carried between sessions. Both are markdown tables so Obsidian renders them and hand edits round-trip, including values containing a literal `|`.

Rules baked in:

- **Only what the user actually told you.** A durable fact about a person is never inferred from an ingested document — that is the document's content and belongs on a page.
- **Confidence is the writer's calibration**, not a measurement.
- **Staleness is reported, never enforced.** An open thread untouched for 30 days is flagged; nothing closes it on the user's behalf.

## Session lifecycle

A new SessionStart hook injects the profile, open threads, and recent activity. Before this, every session started blind and the model had to remember to go read `hot.md`, which it mostly did not.

session lifecycle

**Neither hook can fail a session.** Every error path — no vault configured, package not importable, vault directory missing, slow filesystem — exits 0 with no output. There are tests for each of those paths. The injected block is framed as reference data describing the user, not as instructions.

## Three defects fixed on the way

All three existed on `main` before this branch.

**1. The word cap was documented in five places and enforced in none.**
The second-order bug is the interesting one: the first enforcement attempt counted words of the whole rendered file, but frontmatter plus the generated-file note are ~45 words of fixed boilerplate, so the floor sat within a few words of a tight cap and the trim loop could never get under budget. Fixed by measuring content only.

Trimming needs a defined value order, not arbitrary truncation — activity lines drop first, then contradictions, then threads, and the takeaways truncate last because they are the only part written on purpose.

**2. Deleting the last page in a category stranded its index section.**
The "which headings are generated" set was derived from the pages currently on disk. No pages in `concepts/` means "Concepts" is not in the set, so the stale section fell through to the preserve branch and was kept as though a human wrote it. The reconciler reported `+0 / -0` while still flagging the file changed. Now derived from the canonical category list and the vault's directories.

**3. `HIVEMIND_CAPTURE` was a no-op kill switch.**
`wiki-setup` has long told users to set `HIVEMIND_CAPTURE=false` to skip the end-of-session capture nudge. That variable appears nowhere else in the repo — the Stop hook never read it, so the documented escape hatch did nothing. Implemented as `WIKI_STOP_CAPTURE` to match the neighbouring `WIKI_STOP_REARM_*` knobs, with the documented spelling kept as an alias so no one following the old docs breaks.

## CI gate

`--check` reports drift and exits **2** without writing, so a job can fail on a stale index without a bot committing to the vault. Exit 1 stays reserved for bad input.

memory index --check as a CI gate

## Testing

**70 new tests.** Suite at **915 passed, 4 skipped**.

Coverage is aimed at the guarantees prose could not make:

- **Concurrency** — 60 threads appending log lines, asserting all 60 land and none are lost. This is the regression the lock exists for.
- **Word cap** — a 200-word cap holds against 50 log entries and a 400-word takeaway block; content-vs-file measurement is asserted directly.
- **Index preservation** — a hand-written section survives a rebuild that also drops a deleted page; an emptied category's stale section is removed.
- **Table round-trip** — profile and todo values containing a literal `|`, plus hand edits made in Obsidian.
- **Hook safety** — every failure path of the SessionStart hook asserted to exit 0 silently, plus the opt-out in all four spellings.
- **Doc contract** (`test_memory_docs.py`) — every documented env var is one the code actually reads, every referenced figure ships, and the skills call the CLI instead of restating the procedure. This test is specifically there so defect 3 cannot recur.

### One caveat, stated plainly

`test_stop_hook_behavior.py::test_concurrent_rearm_produces_exactly_one_nudge` was deselected from my full-suite runs. It fails **5/5 on my machine — and 5/5 on a clean tree with this branch stashed**, so it is pre-existing, not a regression from this PR. I verified that by stashing and re-running.

It asserts that ≥3 of 10 rounds demonstrably race two concurrent hook invocations; this host achieves 1. Its own comment says unloaded runs hit 10/10. The assertion is a harness self-check, not a product invariant, and it will fail on any sufficiently loaded or slow-spawn runner. **I did not touch it** — worth a separate issue to gate the ratio on measured spawn latency rather than a fixed 3/10.

## Docs

New **[docs/memory.md](https://github.com/Ar9av/obsidian-wiki/blob/feat/memory-surface/docs/memory.md)** as the concept page, following the `session-brain.md` precedent of a feature page alongside a CLI-reference section.

Updated: `docs/cli.md` (command reference), `docs/architecture.md` (its claim that "the prose files take no lock" was now false), `docs/configuration.md` (new Session hooks section), `.env.example`, `AGENTS.md` (new core principle), and both READMEs — translation parity checked, `check_readme_sync.py` reports clean.

`.skills/llm-wiki/references/MEMORY.md` is the shared procedure the skills point at instead of each carrying its own copy.

## Scope — what this does not do

Four skills are wired to the CLI (`wiki-ingest`, `wiki-update`, `llm-wiki`, plus `wiki-setup` docs). **Eleven still carry their own prose version of the tracking update.** They keep working — nothing is removed — but they should be migrated to `memory sync` in follow-ups. The shared reference exists for exactly that.

Also deliberately out of scope, from the same design pass:

- Persisting the retrieval index (SQLite FTS5) and gating `should_read_precision` in CI
- Stable page IDs with per-claim provenance
- Extending the pre-write git snapshot to the capture and ingest skills
- A Linux systemd timer alongside the macOS launchd plist
- Reconciling the two `.manifest.json` schemas that coexist in the docs

## Review notes

The figures are committed under `docs/images/` and rendered from monospace ASCII, matching the diagram style already used to discuss this design.

Fastest way to exercise it:

```bash
python3 -c "from obsidian_wiki.cli import scaffold_vault; from pathlib import Path; scaffold_vault(Path('/tmp/v'))"
obsidian-wiki memory status --vault /tmp/v
obsidian-wiki memory profile set stack "Python, FastAPI" --confidence 0.85 --vault /tmp/v
obsidian-wiki memory todo add "Try the recap" --vault /tmp/v
obsidian-wiki memory sync --verb INGEST --field source=test.md --vault /tmp/v
obsidian-wiki memory recap --vault /tmp/v
```


Ar9av:
## Update: safe adoption, project scoping, and an agent-facing surface

I reviewed the first commit against a **real 69-page vault** rather than a fixture. It would have degraded that vault in three ways, and a fourth gap made the whole feature a filing cabinet nobody files into. All four are fixed in `1f2c6bd`.

### What the review found

The vault's `index.md` was not a category catalog at all — it was a hand-built "Project Brain" with a projects table, two repo lists, and a Quick Reference section of bash commands.

| | Original commit | Now |
|---|---|---|
| Curated sections | Pushed below 69 generated entries | Kept on top, original order |
| `hot.md` Active Threads narrative | Replaced with "no open threads" | Converted into real todos |
| Huge `WIKI_RESEARCH` log line | Pasted raw, ate the word budget | Summarised and truncated |
| Owner profile | Nothing ever wrote to it | `wiki-capture` populates it |

### Safe adoption

`index.md` and `hot.md` are now written **only** when they carry a `generated_by` marker. Without it the guard refuses, and `memory sync` writes the log line — append-only, always safe — then reports what it skipped rather than failing the ingest.

the migration preview

Migration is **additive, not lossy**. A hand-written `## Active Threads` list is parsed into the todo index *before* the rebuild, so narrative the generator cannot reconstruct survives as live entries. Skipped entirely when todos already exist. A timestamped backup lands in `_archives/` first.

A vault from `setup` is born marked, so this only affects upgrades. `doctor` surfaces the state so nobody has to go looking.

### Closing the loop

This was the real gap. Nothing populated the profile, so the recap had nothing to inject — I had built the storage and the injection but not the extraction.

`wiki-capture` now writes profile facts and open threads, with calibration guidance and one rule that matters most: **a durable fact about a person is never inferred from an ingested document.** That is the document's content and belongs on a page.

### Agent-facing surface

Four MCP tools and six REST endpoints, so a remote agent gets memory rather than just search:

| MCP | REST | Purpose |
|---|---|---|
| `memory_recap` | `GET /v1/memory/recap` | Call at session start; takes `project` |
| `memory_profile` | `GET`/`POST /v1/memory/profile` | `list` / `set` / `forget` |
| `memory_todo` | `GET`/`POST /v1/memory/todos` | `list` / `add` / `done` / `drop` |
| `memory_sync` | `POST /v1/memory/sync` | Reconcile after writes, under one lock |

`memory_sync` returns **409** on an unmigrated vault rather than clobbering it. The server's page write now routes through the shared writer instead of its own ad-hoc log format, so an API write produces the same parseable line as everything else.

The recap is scoped to the project being worked on, derived from the git repo name, so a session about one project is not handed another's threads. It matches a thread's origin *or* its text, and falls back to unscoped rather than implying there is no history.

### Three more bugs, found while building this

- **The marker was never injected into a vault's own preamble.** A migrated vault would have been asked to migrate forever, and sync would have skipped index and hot on every run.
- **Project scoping matched only a todo's origin, not its text**, so the filter missed the obvious case and the fallback then returned everything.
- **`scaffold_vault` did not stamp the marker**, so a brand-new vault was flagged unmigrated and sync skipped the index on every fresh install.

`doctor` also no longer warns on a cosmetic index difference. "Index drift +0/-0" told a reader nothing actionable; it now gates on real added or removed entries.

### Testing

**Suite at 932 passed, 4 skipped.** 102 tests now cover this feature.

New coverage targets the adoption path specifically: the guard refusing a curated vault, dry runs still being allowed, the backup being byte-identical to the original, custom sections keeping their position, narrative threads becoming todos, existing todos never being overwritten, migration being idempotent, and a scaffolded vault being born adopted.

The `test_doctor.py` fixture was updated: it wrote bare stub files, which my check correctly flagged. A real `setup` vault is marked, so the fixture was unrealistic rather than the check being wrong.

The deselected `test_concurrent_rearm_produces_exactly_one_nudge` caveat from the original description still stands — pre-existing, fails 5/5 on a clean tree with this branch stashed.

### Still out of scope

Eleven skills continue to carry their own prose copy of the tracking update. They keep working; they should migrate to `memory sync` in follow-ups. The persisted retrieval index, stable page IDs, the systemd timer, and the two coexisting manifest schemas remain untouched.

Ar9av:
## Update 3: a normal install now actually gets the feature

Reviewing for out-of-the-box UX found the gap that mattered most: **the SessionStart hook was only registered in the repo's own `.claude/settings.json`.** Someone who pip-installs and runs `wiki-setup` got the storage, the CLI, and the MCP tools — and never got memory injected at session start, which is the headline. The only install procedure was prose, and it covered only the Stop hook.

### `obsidian-wiki hooks install`

Hook registration is now code, for both hooks, in `obsidian_wiki/hooks.py`:

```bash
obsidian-wiki hooks install      # idempotent; appends, never clobbers your other hooks
obsidian-wiki hooks status       # registered? bundled? executable? can it reach the package?
obsidian-wiki hooks uninstall    # removes our entries and nothing else
```

- Matches by script name, so a checkout registration at a relative path and a global one at an absolute path don't double up.
- A malformed `settings.json` is refused rather than overwritten.
- `doctor` gained a `session-hooks` check. Unregistered is **info** (the hooks are optional, and a clean `--strict` run must still pass) but the line is always there, because otherwise nobody learns that session-start memory isn't happening. Registered-but-unreachable is a real **warn**.

It's now the third line of the README quickstart, in both languages.

### The silent-failure problem

Both hooks exit 0 on every failure so they can never break a session. The cost is that a missing registration or an unreachable package is invisible from inside one. Three things now answer "why is nothing being injected?":

1. `hooks status` — exit 1 with the reason, no session needed.
2. `doctor` — same check.
3. `WIKI_RECAP_DEBUG=1` — the hook explains each silent exit on stderr. **See the note below; this one is not applied yet.**

### Legacy skills can no longer un-adopt a vault

Adoption is now recorded in `_meta/.memory-adopted`, not only in the per-file marker. Before this, a skill that still rewrote `hot.md` by hand dropped the marker and the next `sync` refused as if the vault had never been migrated. Delete the record to re-arm the guard, e.g. after restoring a curated `index.md` from `_archives/`.

That said, the underlying cause is gone too: **all seventeen write skills now go through `memory sync`**, and no skill tells the model to "create `hot.md` from the template" any more. `wiki-query` and `wiki-narrate` log through `memory log`. The staging CLI's own `STAGE_COMMIT` line, which used a date-only stamp outside the lock, routes through the writer as well. A doc-contract test asserts all of this so it can't regress.

The stale-install nag no longer prints on `memory` and `hooks` commands — `memory` runs from the SessionStart hook and from every write skill, and `hooks` is what fixes the thing the nag is about.

### One thing a person has to do — and one red test until they do

Prismor's `agent-config-tampering` rule blocked two edits, correctly: both touch hook installation.

1. The `WIKI_RECAP_DEBUG` switch in `.claude/hooks/wiki-session-recap.sh`.
2. Replacing the prose hook procedure in `.skills/wiki-setup/SKILL.md` with `obsidian-wiki hooks install`.

I did not route around the block. Both edits are in **`/tmp/blocked-edits.py`** on the box, with anchors asserted so re-running is safe. Run it as yourself, then `pytest tests/test_session_recap_hook.py tests/test_memory_docs.py -q`.

Until that runs, **`test_documented_env_vars_are_the_ones_the_code_reads` fails** — `WIKI_RECAP_DEBUG` is documented in four places and the hook does not read it yet. I extended that test deliberately rather than weaken it: a documented knob nothing reads is precisely the `HIVEMIND_CAPTURE` defect this PR started by fixing, and the guard should catch its own author. Every other test passes.

### Testing

New: `tests/test_hooks.py` (13), adoption-record and re-arm tests in `test_memory.py` (4), doctor coverage for both hook states and both migration states (4), and the extended doc contract (5).

Ar9av:
## Update 4: made it simpler to use

Three rounds of building left a surface that was accurate and unpleasant. Sixteen flags appeared on every `memory` action regardless of relevance, the usage line was a wall of brackets, and the common call needed `--verb` plus one `--field` per value.

### The call that skills make, 17 times over

```bash
# before
obsidian-wiki memory sync --verb INGEST --field source=x.pdf --field pages_created=3

# after
obsidian-wiki memory sync INGEST source=x.pdf pages_created=3
```

`sync` and `log` now take a bare verb and bare `key=value` pairs. `--verb`/`--field` still work and are documented as the older spelling, so nothing breaks.

The worst case, `claude-history-ingest`, went from **seven lines to two**. An unexpected bare word is now a clear error that names the verb it already has, rather than a silent misparse.

**I verified this rather than assuming it.** A script extracts all 27 `obsidian-wiki memory` examples from the skills, fills the placeholders, and executes each against a scratch vault. All 27 run. The one "failure" was `memory todo done <id>` with the placeholder filled as `1`, which correctly reports no such todo.

### `memory --help` leads with examples

```
usage: obsidian-wiki memory <action> [ARGS ...] [options]

examples:
  memory status                                  is the surface current?
  memory sync INGEST source=x.pdf pages=3        log + index + hot, one lock
  memory recap --project myapp                   what a new session should know
  memory profile set stack "Python, FastAPI"     a durable fact about the owner
  memory todo add "Ship the parser"              a thread for the next session
  memory migrate                                 adopt a vault with curated files
  memory index --check                           CI gate; exit 2 on drift
```

Action-specific flags are now in labelled groups — `sync / hot`, `recap`, `profile set`, `todo`, `migrate / compatibility` — so it's obvious which apply to what.

### One less install step

`setup` registers the session hooks and reports the result; `--no-hooks` skips them. The quickstart is two lines again:

```bash
pip install obsidian-wiki
obsidian-wiki setup --vault ~/brain
```

This matters more than it looks. Without the hooks the vault, the CLI, and the MCP tools all work, but no memory is injected at session start — which is the entire point of the feature. Leaving it as a step people had to discover meant most installs would silently lack it.

### Testing

967 passed. New coverage: the positional form, the old flag form still working, the two mixed, a value containing `=`, `sync` with no verb, the error message for a stray bare word, `setup` registering hooks, and `--no-hooks`. Two doc-contract tests now assert that nothing *teaches* `--field` any more and that the quickstart has no separate hook step.

The one expected failure is unchanged: `WIKI_RECAP_DEBUG` stays documented-and-unread until `/tmp/blocked-edits.py` is applied by hand.

Ar9av:
## Update 5: importable, and scoped per user

I looked at what the established memory layers actually expose — **mem0**, **Zep/Graphiti**, **Letta**, and the official **MCP knowledge-graph server** — before changing anything. The gap turned out to be structural.

mem0 is `pip install mem0ai`, then `m.add(messages, user_id="alice")` and `m.search(q, filters={"user_id": "alice"})`. Zep requires a `session_id` on every call. Every one of them is **importable** and **scoped by identity**.

This package had three faces — skills, CLI, MCP — and none of them was importable. You cannot shell out to a CLI on every turn of an agent loop, so in practice it could not be used as agent memory from Python at all.

### `obsidian_wiki.Memory`

```python
from obsidian_wiki import Memory

memory = Memory("~/brain")
memory.remember("stack", "Python, FastAPI", confidence=0.9)
memory.add("Postgres was chosen over MySQL for partial indexes.")
memory.search("postgres")
memory.recap()
```

No API key, no embedding model, no vector store, no network.

### Three verbs, not one — deliberately

mem0 runs an LLM *inside* `add()` to decide what is worth keeping and where it goes. That is the right call when memory is a black box. It is the wrong call here: this package has no LLM, no key and no dependencies, and the whole point of the vault is that a person can read and correct what it stored.

| Call | Stores |
|---|---|
| `add(content)` | knowledge — a thing that is true |
| `remember(key, value)` | a durable fact about the person |
| `todo(text)` | a thread to pick up next time |

The caller already has a model. It does not need a second one hidden in the storage layer guessing which of the three a sentence belongs in.

### Scoping, adapted rather than copied

`user_id` namespaces the **person-shaped** memory into `_meta/profile.<scope>.md`. Pages stay shared.

```python
alice.add("Partial indexes are a Postgres feature.")
bob.search("partial indexes")   # finds it — knowledge is shared
bob.profile()                   # Rust only — memory about a person is not
```

A vault is a shared brain; only what it remembers *about someone* is per-someone. Each scope is a readable markdown table you can open in Obsidian and fix by hand.

The scope reaches the MCP tools and REST routes too. A `user_id` arrives from a request and becomes a filename, so it is validated as untrusted input — hostile values are a **400**, not a filesystem touch. It is a **namespace, not an authorization boundary**, and the docs say so rather than implying isolation it does not provide.

### Three bugs, two of them mine and one embarrassing

- **`search()` returned unfetchable results.** graphrag names the key `page`, not `path`. Every result came back with an empty path until I actually called `get()` on one.
- **Derived titles made 60-character filenames**, because I took the whole opening sentence.
- **The README quickstart didn't work.** It searched `"database choice"` against a page that only ever says "Postgres" and "MySQL". Lexical retrieval has no embedding to bridge that paraphrase, so it returned nothing — my own example demonstrating the known weakness by accident.

On the third: my first doc-test hedged with `search("database choice") or search("postgres")`, which would have let a broken example ship. I fixed the example, documented the constraint, and tightened the test to assert the printed snippet works **as written**.

### Honest positioning

`docs/python-api.md` has a comparison table and says plainly when to use something else:

> **Use a vector layer instead when** you need semantic recall over paraphrased text, you are happy for memory to be opaque, and you already run the infrastructure.
>
> **Use this when** you want to read, grep, edit and `git log` what your agent remembers; when you do not want another database or another API key.
>
> The honest weakness is retrieval: lexical matching misses paraphrase.

### Testing

**999 passed.** `tests/test_client.py` adds 28, including the add→search round trip, OpenAI-message input, scope isolation across profile/todos/recap, pages staying shared, and four hostile `user_id` values rejected before touching disk.

Unchanged: the one expected failure, `WIKI_RECAP_DEBUG`, pending `/tmp/blocked-edits.py` being run by hand.

**Research note:** Prismor blocked `WebFetch` partway through (a floor rule fired on possible prompt injection in fetched content), so this is based on search results plus the official docs I could reach, not full page fetches. Flagging it because it means I could not verify every exact signature first-hand.

Ar9av:
## Regression audit

Asked whether any of this breaks what already worked, I stopped asserting and checked. **It found one real break**, which is now fixed in `7f158af`.

### The break: a venv install registered hooks that did nothing

I installed the built wheel into a clean venv and ran the path a new user takes. Setup reported success. The hooks registered. They did nothing.

A venv, pipx or `uv tool` install puts `obsidian-wiki` somewhere the PATH a hook inherits doesn't include, and the system `python3` can't import the package. Both hooks resolve a runner and **exit 0 silently** when they find none — deliberately, so a hook can never break session startup. That makes this the worst kind of failure: registered, reported as installed, and invisible from inside a session.

`hooks install` now pins the installing interpreter's bin directory onto the front of the registered command's PATH.

My first fix was wrong in a way worth recording: it called `.resolve()` on `sys.executable`, and a venv's `bin/python3` is a symlink to the system interpreter — so resolving handed back `/usr/bin`, pinning the PATH to the one install that *doesn't* have the package. The unresolved path is the venv.

**The existing suite could not have caught this.** It runs from a source tree where the ambient PATH happens to work. Three regression tests now cover it, including the symlink trap.

### What I checked, and what came back clean

| Check | Result |
|---|---|
| Public API removed or renamed | Only `ManifestLockTimeout` / `manifest_lock`, both kept as back-compat wrappers |
| Python 3.9 (CI matrix) | All changed modules have `from __future__ import annotations`; the one PEP 604 pydantic model is in `server.py`, which already requires 3.10 |
| Old CLI spellings (`--verb`/`--field`) | All five forms still exit 0 |
| Skill frontmatter after 18 bulk regex edits | Intact on every file |
| Headings lost in skills | 3 files, all my intentional consolidations, no content lost |
| Wheel contents | 40 skills, both hooks, all new modules |
| `setup.sh` idempotency (CI gate) | Clean tree |
| Every command against a copy of the real 69-page vault | Byte-identical afterwards — read-only commands wrote nothing |
| `lint` / `trust-check` / `eval` non-zero exits | **Identical on `main`** — verified by stashing and re-running, not assumed |
| Staging log format change | Test asserts a substring, unaffected |

### Two false alarms I chased down rather than dismissed

- **"Unbalanced code fences in wiki-dedup."** My checker counted inline ` ``` ` inside prose. `main` has the identical count. Not a regression.
- **`lint` exiting 1 on the real vault.** Looked like breakage; `main` does the same. It means "issues found".

### Pre-existing issue found, deliberately not fixed here

The repo's committed agent-skill mirrors (`.agents/skills/`, `.claude/skills/`, and four others) track **30 of 40 skills**. Ten added later were never committed. Anyone cloning without running `setup.sh` gets a partial set. Present on `main`, unrelated to this PR, worth its own issue.

Running `setup.sh` during the audit generated 57 of those symlinks locally; I removed them so the branch stays focused.

### Behaviour changes an existing user will notice

1. **`setup` now registers the session hooks.** Appends without clobbering; `--no-hooks` opts out.
2. **`STAGE_COMMIT` log lines gain a full ISO timestamp** instead of date-only, now that they go through the shared writer.
3. **A fresh vault gains `_meta/`**, an empty profile and todo table, and an adoption record.

None of these touch an existing vault's contents. The migration guard still refuses to rewrite curated `index.md` or `hot.md` until `memory migrate --apply`.

**Suite: 1002 passed**, with the one expected failure (`WIKI_RECAP_DEBUG`, pending `/tmp/blocked-edits.py`).

